PR #4, opened 2014-11-27
gimp 2.8.14

skela:


skela:
I'll probably not add the non-x11 support myself, so I'll leave that for someone else to play with :)

DomT4:
`GIMP` is now officially part of Homebrew/x11. I made some changes in a follow-up commit which slims down the mandatory dependencies a little and that's something we'll want to keep looking at over time, but as per https://github.com/Homebrew/homebrew-x11/commit/0103bbf734a9d77c6915a282a5e1f39bbc04e0e2 you are now a Homebrew contributor.

Thank you for your patience, I appreciate the amount of work that went into this PR.

skela:
Excellent, thanks @DomT4 !

PR #9, opened 2014-12-28
zim 0.62 (new formula)

DomT4:
New formula for Zim, the “graphical text editor used to maintain a collection of wiki pages”.


zim.rb
```diff
@@ -0,0 +1,57 @@
...
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    %w[simplejson pyxdg].each do |r|
+      resource(r).stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    # Refuses to accept "--single-version-externally-managed"
```
DomT4: Is there any way we can bully the formula into accepting `*Language::Python.setup_install_args(prefix)` here @tdsmith? It gives me an unrecognised option message so I had to revert to the old style of Python handling :cry:

zim.rb
```diff
@@ -0,0 +1,57 @@
...
+  depends_on "cairo"
+  depends_on "gtk+"
+  depends_on "hicolor-icon-theme"
+  depends_on "gtk-engines" => :recommended
+  depends_on "pygtk"
+  depends_on "pygobject"
+  depends_on "py2cairo"
+  depends_on "gobject-introspection"
+  depends_on :x11
+  depends_on :python if MacOS.version <= :snow_leopard
```
tdsmith: Are these all really direct dependencies? Presumably it doesn't link to cairo except through py2cairo, right?

zim.rb
tdsmith: Should go to libexec, shouldn't it?

If it doesn't accept --single-version-externally-managed it doesn't need it, so it's not the end of the world to call setup.py explicitly.

zim.rb
DomT4: Probably not. Hadn't got that far yet - Rewriting a formula I wrote ~6 months ago wasn't pretty :smile_cat:. But nope, no direct Cairo linkage AFAIK.

zim.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  resource "pyxdg" do
+    url "https://pypi.python.org/packages/source/p/pyxdg/pyxdg-0.25.tar.gz"
+    sha256 "81e883e0b9517d624e8b0499eb267b82a815c0b7146d5269f364988ae031279d"
+  end
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    %w[simplejson pyxdg].each do |r|
```
tdsmith: I don't think simplejson is needed on Python > 2.5: https://bazaar.launchpad.net/~jaap.karssenberg/zim/pyzim/view/head:/setup.py#L334

zim.rb
DomT4: It needs to find other elements of `hicolor-icon-theme` and apparently struggles with that when I tuck it into `libexec` because they aren't symlinked into a common prefix with the rest of the `hicolor` stuff.

zim.rb
DomT4: :

zim.rb
```diff
@@ -0,0 +1,57 @@
...
+           "--skip-xdg-cmd", "--record=installed.txt"
+
+    # Directory placement is catered towards Linux rather than OS X.
+    (share/"icons/hicolor/scalable/apps").install_symlink share/"icons/ubuntu-mono-light/apps/22/zim-panel.svg"
+    (share/"pixmaps").install_symlink Dir[share/"zim/pixmaps/*"]
+  end
+
+  def caveats; <<-EOS.undent
+    In your shell profile add:
+      export XDG_DATA_DIRS=#{share}
```
tdsmith: Can we make this unnecessary with env_scripting?

zim.rb
tdsmith: Is this fixable with XDG_DATA_DIRS or is that unrelated? I don't know much about gnomeisms.

DomT4: Unrelated. If you don't set `XDG_DATA_DIRS` the executable won't even start. If you set it but don't move the theme stuff into a common prefix you get runtime warnings that it can't find chunks of the theme.

DomT4:
I added some mildly laughable symlink dancing to deal with the GNOME themes-must-share-common-prefix fun, but the end result is `env` scripting and removal of the caveats. This formula is slowly turning me against including it in `brew` :crying_cat_face:

PR #10, opened 2014-12-29
scantailor: additional boost 1.57 notes

DomT4:
Clarifies the situation around Boost 1.57, with various links to the previous discussions at Homebrew/Homebrew and upstream.

If we don’t see a fix for this within a month or so, it might be worth boneyarding for now.


DomT4:
The test bot failure here was expected, obviously, given the known Boost break.

jkseppan:
I was able to build scantailor with Boost 1.57 with these changes: https://github.com/jkseppan/homebrew-x11/commit/306017778a8cfd9b4cff029ed70fd85d956a1d9f

DomT4:
@jkseppan Fancy submitting a PR for review? :)

PR #12, opened 2015-01-01
deleting tiger-vnc

csabahenk:
The formula is being moved back to
the main homebrew tree.

Cf.
-   http://librelist.com/browser/homebrew/2014/12/31/tiger-vnc-vs-x11/
-   https://github.com/Homebrew/homebrew/pull/35425


DomT4:
It might be good/useful to keep an ` x11 ` variant here?

csabahenk:
It would make sense if someone invested the effort to implement the
vncserver build. It's not a trivial thing though... see the Arch Linux
pkgbuild for reference:
https://projects.archlinux.org/svntogit/community.git/tree/tigervnc/trunk/PKGBUILD

DomT4:
Yeah, having spent some 4 hours re-writing this formula from top to bottom in about 20 different ways, and testing each, there isn't a way forwards to ` vncserver ` until Homebrew allows patching of resources directly, or doesn't frown upon calling ` safe_system "/usr/bin/patch" ` inside ` def install ` there's no way forwards here.

There was previously some discussion of resource patching over at Homebrew/Homebrew at https://github.com/Homebrew/homebrew/issues/31508.

DomT4:
Closing as dead per https://github.com/Homebrew/homebrew/pull/35425, but feel free to resubmit if you'd like to. Thanks for the submission!

PR #13, opened 2015-01-05
nedit 5.6 (new formula)

matlads:


nedit.rb
```diff
@@ -0,0 +1,41 @@
+# Nedit homebrew formnula
```
DomT4: :scissors: this line

nedit.rb
```diff
@@ -0,0 +1,41 @@
...
+class Nedit < Formula
+  homepage "http://www.nedit.org"
+  url "https://downloads.sf.net/project/nedit/nedit-source/nedit-5.6-src.tar.gz"
+  sha1 "133641224b0969e31e3404b874c71c35341f028b"
```
DomT4: `sha256` please

nedit.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  depends_on "lesstif"
+  depends_on :x11
+
+  patch :DATA
```
DomT4: What does this patch do? Is there an upstream bug/patch submission?

nedit.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def install
+    system "make", "macosx", "MOTIFLINK='-lXm'"
+    system "make", "-C", "doc", "man", "doc"
+
+    bin.install "source/nedit"
```
DomT4: No `make install`?

matlads: @DomT4 no, there is no need for a `make install` line. The nedit build produces two binaries only. see the INSTALLATION section here: http://sourceforge.net/projects/nedit/files/nedit-executable/5.5/

nedit.rb
matlads: OK, done.

nedit.rb
matlads: OK, done.

nedit.rb
matlads: This patch makes nedit build against the version of motif available in homebrew.

DomT4: Is our version too new or too old? Has it been submitted upstream?

nedit.rb
DomT4: :+1:

nedit.rb
```diff
@@ -0,0 +1,41 @@
...
+    bin.install "source/nc" => "ncl"
+
+    man1.install "doc/nedit.man" => "nedit.1x"
+    man1.install "doc/nc.man" => "ncl.1x"
```
DomT4: Why `.1x` instead of `.1`?

nedit.rb
```diff
@@ -0,0 +1,41 @@
+class Nedit < Formula
+  homepage "http://sourceforge.net/projects/nedit"
+  url "https://downloads.sf.net/project/nedit/nedit-source/nedit-5.6a-src.tar.gz"
+  sha1 "3f7a39bf0968cf0598bf8d6e2062f243f3aaac96"
```
DomT4: Remove the `sha1` line.

nedit.rb
```diff
@@ -0,0 +1,41 @@
+class Nedit < Formula
+  homepage "http://sourceforge.net/projects/nedit"
+  url "https://downloads.sf.net/project/nedit/nedit-source/nedit-5.6a-src.tar.gz"
```
DomT4: `https://downloads.sourceforge.net/` is preferred over `https://downloads.sf.net`. Tiny nit, I know.

nedit.rb
matlads: .1x indicates that its a man page (section 1) for an X11 application

nedit.rb
matlads: OK. done.

nedit.rb
matlads: OK. Done.

nedit.rb
DomT4: Fair enough. No strong preference either way.

nedit.rb
matlads: LessFif in homebrew is at 0.95.2 which is old (see http://lesstif.sourceforge.net). motif was made open source as openmotif, which is not yet in homebrew.

DomT4: Would we still need the patch if we included `openmotif` in Homebrew first?

DomT4: ?

matlads: We would *not* need this patch if we included `openmotif` in Homebrew. The nedit source has a builtin list of lesstif versions that have been tested and are known to work. This patch adds the version in Homebrew to the list. Therefore, this patch would not be required if `openmotif` were packaged.

matlads: It may be possible to get this patch upstream to the `nedit` development team. I'll see if I can't get on that

DomT4: `openmotif` is available in this tap now.

matlads: I'll rework this patch to use that instead. Updates in the next few days.

DomT4:
@BrewTestBot test this please

dunn:
Just a small thing:
```
==> audit problems
homebrew/x11/nedit:
 * Please remove the indefinite article "A" from the beginning of the description
`1`

**Comment by matlads** [comment]
Done removing the indefinite article "A" from the beginning of the description

File: nedit.rb [hunk 8]
```diff
@@ -0,0 +1,43 @@
+class Nedit < Formula
+  desc "fast, compact Motif/X11 plain text editor"
+  homepage "http://sourceforge.net/projects/nedit"
+  url "https://downloads.sourceforge.net/project/nedit/nedit-source/nedit-5.6a-src.tar.gz"
+  sha256 "53677983cb6c91c5da1fcdcac90f7f9a193f08fa13b7a6330bc9ce21f9461eed"
+
+  depends_on "openmotif"
+  depends_on :x11
+
+  patch :DATA
```
DomT4: Can you add a comment explaining why this is necessary, etc, and ideally a link to an upstream bug report.

bfontaine:
Ping re Dom’s comment above.

nedit.rb
matlads: Nedit specifically checks the version of openmotif that is running against. Unfortunately this check leaves out the latest version of openmotif 2.3.4 ( which is what homebrew has )

see https://sourceforge.net/p/nedit/patches/177/ for the upstream bug report, and patch.

DomT4: Please include that in the formula, just above the `patch` line :)

apjanke:
@BrewTestBot test this please

apjanke:

PR #18, opened 2015-01-27
gkrellm.rb: fixed curry'd inreplace.

shdwjk:
Removed first argument from 3 gsub calls to clear up build issue.


DomT4:
Thanks for the PR! Could you possibly add a `revision 1` line under the initial `sha1`? Thanks! If you wanted to, fixing anything highlighted by `brew audit --strict` would be fantastic, but not mandatory.

PR #23, opened 2015-02-08
easy-tag 2.2.6

robotvert:
Announcement: https://mail.gnome.org/archives/easytag-list/2015-February/msg00003.html

PR #37, opened 2015-03-01
easy-tag 2.2.6

robotvert:
see PR #23 

this new PR simply includes the URL fix from PR #25 so it passes the CI and _hopefully_ gets merged _some day_

(also fix the naming in the commit, the previous PR not respecting the `<name>space<version>` format in the commit message)


DomT4:
@BrewTestBot test this please

DomT4:
@xu-cheng If you get a chance, this should be okay to merge. Bot is :+1: on it.

xu-cheng:
En, downstream bottle job fails. Ping @mikemcquaid for fix.

DomT4:
Rats, It did. Didn't spot that. Happening a lot with taps lately, hitting it with Versions as well.

mikemcquaid:
@xu-cheng You're a maintainer now so when these things work I'd be grateful if you could dig in a bit before pinging.

mikemcquaid:
In this case I've kicked off a rebuild. I'm guessing this is due to Jenkins being restarted or something similar.

xu-cheng:
While, I did kick a rebuild for bottle. It didn't work. (More strangly, although I kicked one rebuild, the Homebrew Bottle tasks run multiple times). Also I noticed the bottles were pushed to the bintray. 

So I was not sure how to fix the problem further. Why didn't I think about rebuilding the whole PR? :smirk:

mikemcquaid:
Fair point, thanks @xu-cheng :bow:

DomT4:
> Also I noticed the bottles were pushed to the bintray.

It *didn't* fail on the ` curl ` step? I've had a few weird situations where it says the downstream bottle build has failed but when you check on Bintray the bottles claim to be available for publishing, and the bottles contain the expected contents, but obviously you don't manually publish them because = problems, and the test-bot just repeats its failure on the bottle build. So the bottles get pushed to Bintray, contain the expected contents, but the bot's ` curl ` publish step fails. Everything seems to work except that publish step, It's odd.

xu-cheng:
@mikemcquaid Bottle task is still failing. And somehow it run for 3 times.

mikemcquaid:
@xu-cheng it auto-retried until I just disabled it.

mikemcquaid:
This might help: https://github.com/Homebrew/homebrew/commit/ed239d1a7e9f70904eb03d3725743e104db384fd

mikemcquaid:
(I've retried the job again)

xu-cheng:
It works. Thanks @mikemcquaid .

@robotvert Thanks for the pull request! 🎉

PR #38, opened 2015-03-09
wcm 0.20.0 (new formula)

corporateshark:
Added the WCM Commander file manager.

### DIFF
--- a/wcm.rb
+++ b/wcm.rb
@@ -1,0 +1,25 @@
+class Wcm < Formula
+  desc "Multi-platform open source orthodox file manager"
+  homepage "http://wcm.linderdaum.com"
+  url "https://github.com/corporateshark/WCMCommander/archive/release-0.20.0.tar.gz"
+  sha256 "f166cfd0ff8511642402892cb8eee3b20c845f82c04ca4ca940996e611aa5aba"
+
+  depends_on "cmake" => :build
+  depends_on "freetype"
+  depends_on "libssh2"
+  depends_on :x11
+  depends_on :macos => :mavericks
+
+  needs :cxx11
+
+  def install
+    ENV.cxx11
+    system "cmake", ".", "-DWITH_SMBCLIENT=OFF", *std_cmake_args
+    system "make", "all"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/wcm", "--help"
+  end
+end


MindTooth:
http://wcm.linderdaum.com/wcm-commander-0-20-0-released/

wcm.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  depends_on "cmake" => :build
+  depends_on "freetype"
+  depends_on "libssh2"
+  depends_on :x11
+  depends_on :macos => :mavericks
+
+  needs :cxx11
+
+  def install
```
DomT4: Should be `ENV.cxx11` under `def install` as well here.

wcm.rb
```diff
@@ -0,0 +1,23 @@
...
+    system "cmake", ".", "-DWITH_SMBCLIENT=OFF", *std_cmake_args
+    system "make", "all", "-j#{ENV.make_jobs}", "-B"
```
DomT4: Specifying `-j` shouldn't be necessary.

wcm.rb
```diff
@@ -0,0 +1,23 @@
+class Wcm < Formula
+  homepage "http://wcm.linderdaum.com"
```
DomT4: Please add a `desc` above the homepage here. Should be less than 80 characters.

wcm.rb
corporateshark: Added

DomT4:
Thanks for the quick response! Could you possibly squash your commits into one? Thanks!

corporateshark:
Done.

DomT4:
Thanks, that seems to have made our test-bot a little happier. Why won't it build on Mountain Lion and below, out of interest?

corporateshark:
I experienced multiple problems with x11/clang configurations of your test-bot. Just disabled those OS X versions to work around those issues.

DomT4:
They may have been `cxx11` failures. I wonder if it'd work without `depends_on :macos => :mavericks` now we've added the `cxx11` ENV.

corporateshark:
I have no idea.

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/658bef8993a47e4b6cc4687062ba7bb53d66f59e. Thanks for your contribution to Homebrew/x11, We appreciate it! :smiley_cat:

PR #39, opened 2015-03-11
swi-prolog: upgrade development version to 7.1.32

honnix:


honnix:
Just checked Jenkins log. Not sure what else needs to be done as it complains about missing dependencies.

mistydemeo:
Thanks!

PR #40, opened 2015-03-14
mikutter 3.2.2 (new formula)

mzyy94:
Create mikutter* formula.

*Mikutter is a multi-pane Twitter client with several advanced features.

(This installation was tested on Yosemite 10.10.1 and 10.10.2.)


mikutter.rb
```diff
@@ -0,0 +1,33 @@
+class Mikutter < Formula
+  homepage "http://mikutter.hachune.net/"
+  url "http://mikutter.hachune.net/bin/mikutter.3.2.2.tar.gz"
+  sha256 "4a8b24a357ccd3132688f7ee97efb909a03a9b78976b19f92e3f50f56dd887ad"
+
+  depends_on "ruby" if /\d\.\d(\.\d)?/.match(`ruby --version 2>&1`).to_s < "1.9.3"
```
mikemcquaid: This should use a `Requirement` instead.

mikutter.rb
```diff
@@ -0,0 +1,33 @@
...
+  depends_on "gtk+"
+  depends_on :x11
+
+  def install
+    if RUBY_VERSION >= "1.9.3"
+      Homebrew.install_gem_setup_path! "bundler" # use gem of system ruby
```
mikemcquaid: We can't install system gems in `install`.

mikutter.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+    open("mikutter.sh", "w+") do |script|
+      script.puts "#!/bin/sh"
+      script.puts "ruby \"#{prefix}/mikutter.rb\" $@"
+    end
+    bin.install "mikutter.sh" => "mikutter"
+  end
+
+  test do
+    system "ruby", "#{prefix}/mikutter.rb", "--help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikutter.rb
mzyy94: First commit of this pull-request:`2f62233731b4fac89ab766786b58040eb9de607c` is written with a `Requirement` but the build bot said `Error': http://bot.brew.sh/job/Homebrew%20X11%20Pull%20Requests/version=yosemite/62/console

I couldn't find any solution. Therefore, I wrote this with if statement.. ;(

mikutter.rb
mzyy94: Ok. I'll simply remove these lines.

mikutter.rb
mzyy94: Mikutter doesn't have substantial options, so I added GUI launching test.
The test is passed on my OS X 10.10.1 and 10.10.2 but on build bot, no.

mikemcquaid:
I'm not sure if this is really going to work in Homebrew, unfortunately. Bundler doesn't really work well for creating independent libraries.

mikutter.rb
```diff
@@ -0,0 +1,40 @@
+class Mikutter < Formula
+  homepage "http://mikutter.hachune.net/"
+  url "http://mikutter.hachune.net/bin/mikutter.3.2.2.tar.gz"
+  sha256 "4a8b24a357ccd3132688f7ee97efb909a03a9b78976b19f92e3f50f56dd887ad"
+
+  depends_on "ruby" if /\d\.\d(\.\d)?/.match(`ruby --version 2>&1`).to_s < "1.9.3"
+  depends_on "gtk+"
+  depends_on :x11
```
mikemcquaid: This should be submitted to `homebrew-x11` instead if it depends on X11.

PR #41, opened 2015-03-20
threadscope 0.2.6 (new formula)

pthariensflame:
This was originally Homebrew/homebrew#37894, but I was told to move it here.

Advice on improving the tests would be greatly appreciated.  :)


threadscope.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  homepage "https://wiki.haskell.org/ThreadScope"
+  url "https://hackage.haskell.org/package/threadscope-0.2.6/threadscope-0.2.6.tar.gz"
+  sha256 "ca3b3a8f57315f47f7f6787e5d92ca26c216fb67562192ae37a8cb37dceecc5f"
+
+  depends_on "pkg-config" => :build
+  depends_on "cabal-install" => :build
+  depends_on "ghc"
+  depends_on "gmp"
+  depends_on :x11
```
mistydemeo: Minor nitpick, but symbol deps go at the top of the list (above build deps and regular deps). The rest of the deps should also be alphabetized within their categories, too.

mistydemeo:
Does threadscope have any headless functionality, or is `--version` the only thing you can do without spinning up a GUI?

pthariensflame:
I don't think it can be used headlessly; its documentation certainly doesn't make mention of it.  There are a couple of "test" flags you can pass, but they all produce a GUI and then leave it running indefinitely.  :/

threadscope.rb
pthariensflame: I will correct this, then.  :)

mistydemeo:
That's fine then, it doesn't sound like there's much more of a test you can do for this software!

pthariensflame:
@mistydemeo Alterations done!

threadscope.rb
```diff
@@ -0,0 +1,39 @@
...
+  sha256 "ca3b3a8f57315f47f7f6787e5d92ca26c216fb67562192ae37a8cb37dceecc5f"
+
+  depends_on :x11
+  depends_on "cabal-install" => :build
+  depends_on "pkg-config" => :build
+  depends_on "cairo"
+  depends_on "fontconfig"
+  depends_on "freetype"
+  depends_on "gettext"
+  depends_on "ghc"
```
mistydemeo: Is this needed at runtime, or is it buildtime-only?

pthariensflame: It's needed at runtime.  ThreadScope is a graphical parallel profiling tool for Haskell, so it needs access to the compiler (and its ABI) to do its job.

mistydemeo:
Great, thanks! Pulled.

PR #47, opened 2015-03-30
Cleanup mupdf formula

ojab:
Style fixes, test & conflict with mupdf-tools added.


mupdf.rb
```diff
@@ -14,12 +12,21 @@ class Mupdf < Formula
 
   depends_on :macos => :snow_leopard
   depends_on :x11
+  depends_on "openssl"
```
DomT4: If this was linking into the system OpenSSL before, ` revision ` the formula.

ojab: thanks, fixed

xu-cheng:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

PR #53, opened 2015-04-10
 gtk-mac-integration 2.0.5 (new formula)

yesimon:
Formula for gtk-mac-integration.

Redirected here from Homebrew/homebrew#38470.


tschoonj:
This formula was added to homebrew core with Homebrew/homebrew#39868 and can be closed

PR #35, opened 2015-03-01
revision bumps to use the latest poppler

bfontaine:
Depends on Homebrew/homebrew#37280. (WIP)

* ~~`pdf2svg`~~
* `xournal`
* ~~`inkscape` (doesn’t build, even without poppler, due to a `libiconv` error)~~

PR #59, opened 2015-04-16
inkscape 0.91

dunn:
Figured out what was keeping inkscape from compiling in #35—just had to add some flags to make it pick up the system iconv.

If this works then Homebrew/homebrew#37280 can finally move ahead.


dunn:
:confetti_ball: should this not be merged until after the poppler stuff goes through?

inkscape.rb
```diff
@@ -10,83 +10,61 @@ class Inkscape < Formula
...
-  end if MacOS.version < :mavericks
+  if MacOS.version < :mavericks
+    fails_with :clang do
+      cause "inkscape's dependencies will be built with libstdc++ and fail to link."
+    end
+  end
 
   def install
     ENV.cxx11
+    ENV.append_to_cflags "-I/usr/include"
```
dunn: Don't think this line is actually necessary but is it worth making the bot rebuild for that?

DomT4:
@BrewTestBot test this please

DomT4:
If this passes the bot I'll merge it. We've been trying to bump `poppler` for 2+ months now, and keep getting hung up. People keep expressing frustrating with the Inkscape failure to build in the mean time, understandably. If we merge this and then a week later it needs revisioning for `poppler`, so be it.

DomT4:
Massively died on [Mountain Lion](http://bot.brew.sh/job/Homebrew%20X11%20Pull%20Requests/114/version=mountain_lion/consoleFull) sadly. Passed the other two bots fine.

dunn:
Very weird, since it passed when I first opened this PR.

DomT4:
It did indeed. The only major thing I can think that's changed in the mean time is `boost`. I don't think anything else in the deps list has had a major update.

dunn:
And GCC 5 instead of 4.9; could that make a difference?

DomT4:
Possibly. I'd be surprised if some formulae were having ML compile problems and we hadn't yet been hit with a :mount_fuji: of issues though. Are there any recent bug reports upstream on `gcc` compiles?

dunn:
Doesn't look like it, so that's probably not the issue.

dunn:
Does this https://bugs.launchpad.net/inkscape/+bug/1293295/comments/5 mean that we should try `ENV.cxx11 if MacOS.version >= :mavericks`?

DomT4:
Interesting bug comment. Give it a shot, see what happens. You may need to change the `needs` line as well.

dunn:
Yeah, they both needed changing. But it worked!

inkscape.rb
```diff
@@ -10,83 +10,60 @@ class Inkscape < Formula
...
   depends_on :x11
-  depends_on 'poppler' => :optional
-  depends_on 'hicolor-icon-theme'
-
-  needs :cxx11
 
-  fails_with :clang do
-    cause "inkscape's dependencies will be built with libstdc++ and fail to link."
-  end if MacOS.version < :mavericks
+  if MacOS.version < :mavericks
```
DomT4: I'm not sure this needs to be conditional. I think we can just keep the

```ruby
if MacOS.version < :mavericks
  fails_with :clang do
    cause "inkscape's dependencies will be built with libstdc++ and fail to link."
  end
end
```

Bit and then tack `ENV.cxx11 if MacOS.version >= :mavericks` into `def install` as it currently is.

dunn: `needs :cxx11` does need to be conditional, but that probably wasn't the most natural way to write it.

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/31c0fb91e7de69808a7af774f49207fe06df8f61. Not 110% certain on the syntax around sometimes mandatory `cxx11` but it passed the bot and we can push minor fixes to the formula if any become necessary. Thanks for sorting this out @dunn! :trophy: :cake: :fireworks:

PR #60, opened 2015-04-21
gv 3.7.4

bfontaine:
See Homebrew/homebrew#38914


mikemcquaid:
Thanks again @bfontaine!

PR #64, opened 2015-04-28
gnumeric 1.12.22 Closes Homebrew/homebrew#39096

tschoonj:
Version bump
Formula cleanup
This version does not work on my system though: I get a bus error
immediately on startup which looks like an upstream issue.
Also: this formula no longer belongs in this tap since it depends on
gtk3, which now only has a quartz backend (see Homebrew/homebrew#38972)


geoffreak:
I can also verify this causes a bus error: `Bus error: 10`

jacknagel:
this formula was updated.

PR #72, opened 2015-05-15
gnumeric 1.12.22

mistydemeo:
Fixes Homebrew/homebrew#39764.


tschoonj:
@mistydemeo I already had an open pull request for this one from a couple of weeks ago (#64). Yours doesn't deal properly with the schemes file: you shouldn't delete it, just stop it from being generated by `make` using the configure option `--disable-schemas-compile`, followed by a post_install step that will produce the correct file based on all formulas that support it. Your solution will probably impact correct functioning of gnumeric.

Also, this version doesn't appear to work at all as I already mentioned in my PR: it segfaults immediately on startup due to some upstream issue.

Keep in mind that this formula should be put back in the main Homebrew tap since it doesn't rely on X11 anymore due to gtk+3 only having the quartz backend since 3.16.0

mistydemeo:
Sorry about that - I missed your earlier PR.

It looks like this crash is coming from goffice, within the `go_quad_constant8` function.

mistydemeo:
I've found a fix for the goffice fix; will apply tomorrow.

mistydemeo:
`brew update` and reinstall goffice, and this should work now. Also adapted your commit from #64 for the new version.

tschoonj:
Thanks!


Dr. Tom Schoonjans
Research Fellow

Bundesanstalt für Materialforschung und -prüfung (Federal Institute for Materials Research and Testing)
Richard-Willstätter-Strasse 11, 12489 Berlin
Germany

Telephone +49 (0) 30-8104-5612

Tom.Schoonjans@bam.de<mailto:Tom.Schoonjans@bam.de>
www.bam.de


On 16 May 2015, at 18:01, Misty De Meo <notifications@github.com<mailto:notifications@github.com>> wrote:


brew update and reinstall goffice, and this should work now. Also adapted your commit from #64<https://github.com/Homebrew/homebrew-x11/pull/64> for the new version.

PR #77, opened 2015-05-24
pgplot: add mirrors/modernise

DomT4:
Combination of two commits.


DomT4:
`snid` failure is unrelated. It's a Homebrew/science formula which needs bumping after the `gcc` update. They're aware but I've dropped them a notification. Since this isn't a bump and won't force an immediate rebuild for anyone, merging.

PR #78, opened 2015-05-25
stable 7.2.0 and dev 7.3.1

honnix:
- Update swi-prolog to stable version 7.2.0 and development version 7.3.1
- Change from sha1 to sha256
- Temporarily comment out bottle as there is no existing bottle for 7.2.0


DomT4:
> Temporarily comment out bottle as there is no existing bottle for 7.2.0

You can leave this as it was. It'll be automatically updated when I pull it later.

honnix:
@DomT4 OK, will fix it.

PR #80, opened 2015-06-08
libxkbcommon 0.5.0 (new formula)

afh:
libxkbcommon is a library to handle keyboard descriptions, including
loading them from disk, parsing them and handling their state.

### DIFF
--- a/libxkbcommon.rb
+++ b/libxkbcommon.rb
@@ -1,0 +1,41 @@
+class Libxkbcommon < Formula
+  desc "Keyboard handling library"
+  homepage "http://xkbcommon.org/"
+  url "http://xkbcommon.org/download/libxkbcommon-0.5.0.tar.xz"
+  sha256 "90bd7824742b9a6f52a6cf80e2cadd6f5349cf600a358d08260772615b89d19c"
+  head "https://github.com/xkbcommon/libxkbcommon.git"
+
+  depends_on :x11
+  depends_on "bison" => :build
+  depends_on "pkg-config" => :build
+
+  def install
+    inreplace "configure" do |s|
+      s.gsub! "-version-script $output_objdir/$libname.ver", ""
+      s.gsub! "${wl}-version-script", ""
+    end
+    inreplace %w[Makefile.in Makefile.am] do |s|
+      s.gsub! "-Wl,--version-script=${srcdir}/xkbcommon.map", ""
+      s.gsub! "-Wl,--version-script=${srcdir}/xkbcommon-x11.map", ""
+    end
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+    #include <stdlib.h>
+    #include <xkbcommon/xkbcommon.h>
+    int main() {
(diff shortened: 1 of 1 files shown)


libxkbcommon.rb
```diff
@@ -0,0 +1,29 @@
+class Libxkbcommon < Formula
+  desc "Keyboard handling library"
+  homepage "http://xkbcommon.org/"
+  url "http://xkbcommon.org/download/libxkbcommon-0.5.0.tar.xz"
+  sha256 "90bd7824742b9a6f52a6cf80e2cadd6f5349cf600a358d08260772615b89d19c"
+  head "https://github.com/xkbcommon/libxkbcommon.git"
+
+  depends_on :x11
```
DomT4: If this absolutely depends on `x11` in a non-optional way it should be filed at Homebrew-x11.

DomT4: Ha, you *did* file it at Homebrew-x11. Ignore me please, I can't read repository names apparently :smile_cat:

afh: You had me worried there for a minute :laughing:

libxkbcommon.rb
```diff
@@ -0,0 +1,29 @@
...
+  depends_on "bison" => :build
```
DomT4: Will this not build against the system `bison`?

libxkbcommon.rb
DomT4: Sorry, long week, I'm not entirely reliable at this point :wink:.

libxkbcommon.rb
afh: No, unfortunately not, due to a syntax error in `/src/xkbcomp/parser.y` which does not occur when Homebrew's bison is used.

```
==> make install
/bin/sh ./build-aux/ylwrap src/xkbcomp/parser.y y.tab.c src/xkbcomp/parser.c y.tab.h `echo src/xkbcomp/parser.c | sed -e s/cc$/hh/ -e s/cpp$/hpp/ -e s/cxx$/hxx/ -e s/c++$/h++/ -e s/c$/h/` y.output src/xkbcomp/parser.output -- bison -y -d -p _xkbcommon_ 
/private/tmp/libxkbcommon20150609-80633-l2iipq/libxkbcommon-0.5.0/src/xkbcomp/parser.y:220.5-9: syntax error, unexpected type, expecting string or identifier
make: *** [src/xkbcomp/parser.c] Error 1
```

DomT4: Potentially worth reporting upstream to see if they can fix it, but I'm happy enough to carry the dep.

DomT4:
Is there any chance we can add a test here?

afh:
@DomT4 If compiling a simple program (similar to how `configure`, `cmake`, et al often do it) is acceptable, then yes (you wouldn't happen to know a formula that does this from the top of your head by any chance, would you?).

Otherwise I wouldn't know how to test `libxkbcommon`, since it does not provide any executable programs.

DomT4:
If it's only a library you can write a test [like this one](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/tinyxml2.rb#L22-L32).

afh:
Thanks for the link to the example, it was quite helpful to add the test for this formula.

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/487be2685d534cad22833751e4f21f18fc8bc46a. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #82, opened 2015-06-09
A useful task timer.

vurte:
I was using this task timer on Linux. But when I switched back to Mac I didn't found a free 
task timer with the same features like gtimer.
It has good reporting features and the storage formate is in plain text.

I think it is worthy to be added

ps: I added gtimer first on homebrew but because of the x11 dependency I moved it to homebrew-x11.
https://github.com/Homebrew/homebrew/pull/40524#discussion_r32024430

### DIFF
--- a/gtimer.rb
+++ b/gtimer.rb
@@ -1,0 +1,82 @@
+class Gtimer < Formula
+  desc "An application for timing how you spend your time among projects and tasks."
+  homepage "http://www.k5n.us/gtimer.php"
+  url "https://downloads.sourceforge.net/project/gtimer/2.0.0/gtimer-2.0.0.tar.gz"
+  sha256 "a2dd70793a5b1b4d497be0c4429b8e8cef6910497b6af8fbdedff00764308458"
+
+  depends_on :x11
+  depends_on "gtk+"
+
+  patch :DATA
+
+  def install
+    ENV.O1
+    system "./configure", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+                          "--with-gtk2"
+    system "make"
+    system "make", "install"
+  end
+
+end
+
+__END__
+diff --git a/task.c b/task.c
+index baec803..8067369 100644
+--- a/task.c
++++ b/task.c
+@@ -56,7 +56,7 @@
+ #include <dirent.h>
+ #endif
+ #include <errno.h>
+-#include <malloc.h>
++#include <malloc/malloc.h>
+ #include <string.h>
+ #include <ctype.h>
+ #include <sys/types.h>
+
+diff --git a/project.c b/project.c
+index 5094d13..aaf8f31 100644
+--- a/project.c
++++ b/project.c
+@@ -40,7 +40,7 @@
+ #include <dirent.h>
+ #endif
+ #include <errno.h>
+-#include <malloc.h>
++#include <malloc/malloc.h>
+ #include <string.h>
+ #include <ctype.h>
+ #include <sys/types.h>
+
+diff --git a/custom-list.c b/custom-list.c
+index 1e83c18..31b7606 100644
+--- a/custom-list.c
++++ b/custom-list.c
+@@ -610,8 +610,8 @@ custom_list_append_record (CustomList   *custom_list,
+   gulong        newsize;
+   guint         pos;
+ 
+-  g_return_if_fail (CUSTOM_IS_LIST(custom_list));
+-  g_return_if_fail (name != NULL);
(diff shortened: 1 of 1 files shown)


vurte:
Option -> Idle Detection  should be deactivated in Mac.

DomT4:
> "Options -> Idle Detection" should be deactivated on Mac.

Is this part of the formula/something we can do/warn about?

vurte:
Normally on Linux Idle Detection warns you if you not working at the pc (no mouse and keyboard action) and you can revert,resume or continue the task.
On my Mac I figured that it shows always an  idle Message after 5 Minutes even if I'm working.
So best is not to use it and deactivate the feature. GTimer gives also manual control. So you can remove the time from the task if you become idle and did paused the task before.

gtimer.rb
```diff
@@ -0,0 +1,81 @@
...
+  patch :DATA
+
+  def install
+    ENV.O1
+    system "./configure", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+  test do
```
DomT4: Let's just scrap the test. It's not doing anything useful, and I can manually ignore the bot failure if the test is the only thing missing.

gtimer.rb
```diff
@@ -0,0 +1,81 @@
...
+  url "https://downloads.sourceforge.net/project/gtimer/2.0.0/gtimer-2.0.0.tar.gz"
+  sha256 "a2dd70793a5b1b4d497be0c4429b8e8cef6910497b6af8fbdedff00764308458"
+
+  depends_on :x11
+  depends_on "gtk+"
+
+  patch :DATA
+
+  def install
+    ENV.O1
```
DomT4: Is this a problem on Linux as well?

`ENV.deparallelize` is preferred.

gtimer.rb
```diff
@@ -0,0 +1,81 @@
+class Gtimer < Formula
+  desc "An application for timing projects and tasks."
+  homepage "http://www.k5n.us/gtimer.php"
+  url "https://downloads.sourceforge.net/project/gtimer/2.0.0/gtimer-2.0.0.tar.gz"
+  sha256 "a2dd70793a5b1b4d497be0c4429b8e8cef6910497b6af8fbdedff00764308458"
+
+  depends_on :x11
+  depends_on "gtk+"
+
+  patch :DATA
```
DomT4: Has this patch been reported upstream? If so, can you add a link to that bug report/patch submission? If not, can you submit it upstream and then add the link, Thanks!

DomT4:
> No it is not part of the formula. It's about the behavior on Mac.

Right. If this is an internal setting that needs tweaking, feel free to add a couple of lines as a caveat about the feature being incompatible with OS X and should be disabled, etc.

gtimer.rb
vurte: When I remove ENV.O1 it also compiles but the application exits with "Segmentation fault: 11". 

I will check if I can use ENV.deparallelize.

DomT4: Can you report the segfault upstream? It'd be good to get that fixed.

vurte: I don't exactly  know what "ENV.O1" does, but with it there is no  "Segmentation fault: 11".

gtimer.rb
vurte: OK I will report it. I mailed the developer before, but It would be better in sourceforge.

gtimer.rb
DomT4: Oh, Sorry - I got my mental wires crossed. No need to replace with `ENV.deparallelize`. Reporting it upstream would be good though. Something as simple as a bug report upstream with "gtimer segfaults on OS X when built with Clang unless O1 is passed" would work.

vurte: OK thanks I will do it :)

gtimer.rb
vurte: Testing would be possible, if there where a x11 server (XQuartz)  running in the background to use it fore the tests.  Do you have this on the roadmap ?

DomT4: The test-bot has X11 available for build and testing, so you can use it. You can kill it afterwards by doing something [like this](https://github.com/Homebrew/homebrew-x11/blob/master/rxvt-unicode.rb#L48-L55) if it doesn't exit quickly enough for the test.

vurte: this is :+1:

gtimer.rb
bfontaine: Anything new here?

apjanke:
Closing due to lack of response.

PR #83, opened 2015-06-12
Gtk+ goes quartz only

tschoonj:
This branch continues my work in Homebrew/homebrew#39868 and Homebrew/homebrew-science#2311 regarding removing X11 support in Gtk+ and its dependencies cairo and pango in favor of quartz.

I found the following packages in this tap to have a mandatory or recommended dependency on gtk+:

- [x] bochs
- [x] easy-tag
- [x] eiffelstudio (bug triggered: `gdkimage-quartz.c:325:_gdk_image_new_for_depth: assertion failed: (depth == 24 || depth == 32)`)
- [x] fsv -> boneyard
- [x] geany
- [x] geda-gaf
- [x] geeqie
- [x] gerbv
- [x] ggobi
- [x] gkrellm -> hard X11 dependency through gdkx -> boneyard
- [x] glade
- [x] gobby
- [x] gsmartcontrol
- [x] gtk-chtheme
- [x] gtksourceviewmm
- [x] gtksourceviewmm3 (needs regular update)
- [x] gtkwave
- [x] hexchat
- [x] inkscape
- [x] iptux
- [x] kernagic
- [x] klavaro
- [x] meld
- [x] pcb
- [x] pdf2svg
- [x] prooftree
- [x] sylpheed
- [x] terminator
- [x] threadscope
- [x] upnp-router-control
- [x] viewglob -> hard X11 dependency -> boneyard
- [x] xchat
- [x] xournal
- [x] yarp

This branch should not be merged in until:

  1. Homebrew/homebrew#39868 is merged in
  2. the bottles in this branch have been rebuilt against those that were created for Homebrew/homebrew#39868

Several formulas will have a failing build due to a dependency on gtk-mac-integration, which I added as a new formula in Homebrew/homebrew#39868.
Most formulas that will be modified in this PR will eventually have to be moved out of this tap since they will no longer depend on X11. This is also true for all formulas here depending on `gtk+3`


tschoonj:
@mikemcquaid ok, I updated the patch descriptions for the X11 tap and submitted upstream tickets where possible. Will push when the main PR is merged in. I did the same for the Homebrew/homebrew-games#268 PR: these can already be merged in since I managed to move them all to gtk+3 support.

No patch comments were necessary for the homebrew-science tap in Homebrew/homebrew-science#2311

gtksourceviewmm.rb
```diff
@@ -1,17 +1,120 @@
...
+  sha256 "0000df1b582d7be2e412020c5d748f21c0e6e5074c6b2ca8529985e70479375b"
 
-  depends_on 'gtksourceview'
-  depends_on 'pkg-config' => :build
-  depends_on 'gtkmm'
-  depends_on :x11
+  depends_on "gtksourceview"
+  depends_on "pkg-config" => :build
+  depends_on "gtkmm"
+  revision 1
```
DomT4: The `revision 1` line here is misplaced. You can either fix it just after the core PR is merged (Thus triggering another build) or I'll move it up after merge.

inkscape.rb
```diff
@@ -33,7 +35,6 @@ class Inkscape < Formula
   depends_on "little-cms"
   depends_on "pango"
   depends_on "popt"
-  depends_on :x11
```
DomT4: I thought Inkscape's `x11` dep was independent of GTK for some reason? I thought Inkscape actually used x11 for their GUI?

pdf2svg.rb
```diff
@@ -1,20 +1,60 @@
...
+    touch("README")
+    system "autoreconf", "-i", "-f"
+    system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    curl "-O", "http://partners.adobe.com/public/developer/en/xml/AdobeXMLFormsSamples.pdf"
```
DomT4: We should vendor this as a `resource` instead of using `curl`.

tschoonj:
inkscape uses X11 for their Mac app bundles but it also works with the quartz backend. The main issue is that it doesn't use the native Mac menubar since they didn't include gtk-mac-integration support. 

Hope the developers will fix that one day (or move to gtk+3...)

tschoonj:
By the way here there are three formulas for the boneyard: `fsv`, `gkrellm` and `viewglob`. There is also a problem with `Eiffelstudio`: it compiles and installs fine, but the graphical user interface crashes on startup due to some issue in the gtk-quartz backend. This will not work until the Eiffelstudio people move to gtk+3, so in the meantime only the command-line tools are useable.

DomT4:
> inkscape uses X11 for their Mac app bundles but it also works with the quartz backend. The main issue is that it doesn't use the native Mac menubar since they didn't include gtk-mac-integration support.

Is the GUI still available in some format?

> By the way here there are three formulas for the boneyard: fsv, gkrellm and viewglob.

Tee up a PR at the Boneyard to remove them, and add them to this PR as removals.

> but the graphical user interface crashes on startup due to some issue in the gtk-quartz backend

Is the GUI a separate thing? Can we remove it without removing the command-line tools?

tschoonj:
sure running `inkscape` will give you the GUI like before but it will be running using gtk's quartz backend. At least it will look better than the x11 version on retina Macs.

Will remove the three formulas in a commit, which I will push as soon as the commit in homebrew is merged in. This will kick off the new build then.

About Eiffelstudio: I noticed that the GUI is not built when the gtk+ dependency is removed from the formula. You will just get the command-line tools then.

DomT4:
> sure running inkscape will give you the GUI like before but it will be running using gtk's quartz backend. 

Cool. Sounds good :+1:.

> Will remove the three formulas in a commit, which I will push as soon as the commit in homebrew is merged in. 

:+1:. If you could fix the few other nits I highlighted at the same time that'd be awesome, Thanks!

> I noticed that the GUI is not built when the gtk+ dependency is removed from the formula. 

As long as there aren't any complications to CLI usage, feel free to kill the `gtk+` dep.

mistydemeo:
Ping @manuseiffel for Eiffel discussion

kt10aan:
Yarp has moved away from gtk and is using qt5. Also, a lot of other stuff changed in yarp, so the changes in your PR are superceded by  PR #85. So, yarp.rb should be taken out of your PR.

tschoonj:
@kt10aan ok thanks for the heads-up. Will take care of it.

kt10aan:
@tschoonj, thanks!

tschoonj:
@mikemcquaid the x11 build failed but I don't understand why. Could you have a look at this? Thanks

mikemcquaid:
@tschoonj It hasn't run yet.

mikemcquaid:
@BrewTestBot test this please

tschoonj:
@mikemcquaid things look stuck on the buildbot... something went wrong?

mikemcquaid:
@tschoonj There's a lot going on, please be patient.

DomT4:
Failures were:

* `orbit` deciding to fail to compile:
```
----------------------------------------------------------------------
Making install in inhibit
(rm -f baa.h baa-stubs.c baa-skels.c baa-common.c baa-imodule.c baa-skelimpl.c || true) > /dev/null
(rm -f foo.h foo-stubs.c foo-skels.c foo-common.c foo-imodule.c foo-skelimpl.c || true) > /dev/null
../../src/idl-compiler/orbit-idl-2 --showcpperrors -I. --deps ./.deps/baa.idl.P baa.idl
../../src/idl-compiler/orbit-idl-2 --showcpperrors -I. --deps ./.deps/foo.idl.P foo.idl
orbit-idl-2 2.14.19 compiling
  mode, show preprocessor errors, passes: stubs skels common headers 

Processing file baa.idl
orbit-idl-2 2.14.19 compiling
  mode, show preprocessor errors, passes: stubs skels common headers 

Processing file foo.idl

** (orbit-idl-2:24508): WARNING **: failed to create '.deps' directory 'File exists'
for I in foo.h foo-stubs.c foo-skels.c foo-common.c foo-imodule.c foo-skelimpl.c; do \
		if test -f $I; then \
			sed -e 's,OObject,Object,g' -e 's,TTypeCode,TypeCode,g' $I > $I.out; \
			mv $I.out $I ; \
		fi; \
	done
true
make[3]: *** [baa.h] Segmentation fault: 11
make[3]: *** Deleting file `baa.h'
make[2]: *** [install-recursive] Error 1
make[1]: *** [install] Error 2
make: *** [install-recursive] Error 1
==> Formula
Path: /usr/local/Library/Formula/orbit.rb
```

That's probably a race condition - It didn't persist across all three bots.

* Meld failed to compile:

```
msgfmt help/cs/cs.po -o build/help/cs/cs.mo
itstool -m build/help/cs/cs.mo -o build/help/cs help/C/command-line.page
Traceback (most recent call last):
  File "/usr/local/opt/itstool/bin/itstool", line 25, in <module>
    import libxml2
ImportError: No module named libxml2
error: command 'itstool' failed with exit status 1
==> Formula
Tap: homebrew/homebrew-x11
Path: /usr/local/Library/Taps/homebrew/homebrew-x11/meld.rb
```

* Prooftree failed in a strange way:

```
==> ./configure --prefix /usr/local/Cellar/prooftree/0.12
ocamlc version 4.02.1 found.
ocamlopt.opt version 4.02.1 found. Native compilation enabled.
test ocamlopt.opt -I +lablgtk2
File "_none_", line 1:
Error: Cannot find file gtkInit.cmx
./configure: line 132: ocamlfind: command not found
./configure: line 141: [: =: unary operator expected

  Configuration summary:
    binaries   will be copied to /usr/local/Cellar/prooftree/0.12/bin
    man pages  will be copied to /usr/local/Cellar/prooftree/0.12/share/man
    native-code compilation enabled with ocamlopt.opt
    LablGtk2 at 
==> make -j1
Makefile:107: Makefile.deps: No such file or directory
```

Everything else was fine.

tschoonj:
The `kernagic` failure is weird: Homebrew appears to add `-fpeel-loops` to the clang argument list which is only supported on the version that ships with Yosemite

mikemcquaid:
Merged in d612245. I've merged everything except changes to `kernagic`, `meld` and `prooftree`. Can you open a new PR to work on them? It'll be much easier for the bot to just have three formulae to test too.

tschoonj:
Thanks!

Will try and fix the other formulas

manuseiffel:
Sorry, for the late reply, that discussion did not reach me back then. I noticed it because I just recently updated the formulae to use the latest release of EiffelSTudio.

Currently EiffelStudio does not support GTK+3 and it GTK quartz crashes on an assertion which doesn't make sense. I see no reason why it would fail on GTK quartz when it works just fine on GTK X. Can we still have EiffelStudio compile against the X version of GTK until the GTK quartz people solve the issue? It prevents anyone from using EiffelStudio.

tschoonj:
I think this was fixed with a patch to gtk+ that was added in Homebrew/homebrew#41339.

Though the patch has been submitted to the corresponding [bugzilla report](https://bugzilla.gnome.org/show_bug.cgi?id=557780), it seems that they are still waiting for someone to test it properly before merging it in.

I haven't checked it but this patch should fix the issue with EiffelStudio.

The gtk+ bug is in the quartz backend, so it would not necessarily occur when running with other backends such as X11 and win32.

manuseiffel:
Good to know. How can we help to get that patch tested and move upstream?

tschoonj:
If you can confirm that this patch fixes the problem with EiffelStudio and leave a note on the bugzilla thread, I am sure it will help a lot.

Issue #24, opened 2015-02-09
mit-scheme does not work


Hardmath123:
See homebrew/homebrew#33375 (fix in commit https://github.com/Homebrew/homebrew/commit/8c30f0bb124d5b5ecee132649b17b98df5cb4f79 but not in this repo), affects file https://github.com/homebrew/homebrew-x11/blob/master/mit-scheme.rb.

DomT4:
Can you clarify? The `inreplace` fix you pointed to is included in the X11 formula here.

Hardmath123:
Hmm… that's strange. I ran `brew update` and cleared the cache, but I still get the following error:
```
$ brew install mit-scheme
==> Installing mit-scheme from homebrew/homebrew-x11
==> Downloading http://ftpmirror.gnu.org/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz
######################################################################## 100.0%
Error: inreplace failed
lib/include/configure:
  expected replacement of "/usr/local" with #<Pathname:/…shortened…/homebrew/Cellar/mit-scheme/9.2>
```
(Yeah, my `brew` isn't in `/usr/local`, but that shouldn't be an issue?)

DomT4:
It shouldn't be, but Ruby seems to think it is. Anything show up on `brew doctor`?

Hardmath123:
Outdated XQuartz is the only thing that seems relevant, but I doubt this has anything to do with that.

DomT4:
Would be surprised. What is your prefix? `brew config` will show it and a bunch of other details that could be useful. Is it a real prefix or a symlink elsewhere?

Hardmath123:
```
$ brew config
HOMEBREW_VERSION: 0.9.5
ORIGIN: https://github.com/Homebrew/homebrew.git
HEAD: 1f2c60618edc2f947fbc2ca8d1f9378d2b428b2b
Last commit: 12 hours ago
HOMEBREW_PREFIX: [~]/my_bin/homebrew
HOMEBREW_CELLAR: [~]/my_bin/homebrew/Cellar
CPU: quad-core 64-bit ivybridge
OS X: 10.9.5-x86_64
Xcode: 5.1.1
CLT: N/A
Clang: 5.1 build 503
X11: 2.7.6 => /opt/X11
System Ruby: 2.0.0-p481
Perl: /usr/bin/perl
Python: /usr/bin/python
Ruby: /usr/bin/ruby
[...]

PR #84, opened 2015-06-14
mit-scheme: lots of formula fixups

L2G:
* Use ENV.deparallelize instead of ENV.j1
* Don't patch src/lib/include/configure
* Clean up inreplace usage
* Replace certain "http:" URLs with "https:"
* Make x11 dependency "recommended"
* Add test based on a prime generator
* Rewrap some over-wide comment blocks
* Other changes to pass `brew audit --strict`

### DIFF
--- a/mit-scheme.rb
+++ b/mit-scheme.rb
@@ -1,10 +1,10 @@
-require "formula"
-
 class MitScheme < Formula
-  homepage "http://www.gnu.org/software/mit-scheme/"
+  desc "MIT/GNU Scheme development tools and runtime library"
+  homepage "https://www.gnu.org/software/mit-scheme/"
   url "http://ftpmirror.gnu.org/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
-  mirror "http://ftp.gnu.org/gnu/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
-  sha1 "d2820ee76da109d370535fec6e19910a673aa7ee"
+  mirror "https://ftp.gnu.org/gnu/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
+  sha256 "4f6a16f9c7d4b4b7bb3aa53ef523cad39b54ae1eaa3ab3205930b6a87759b170"
+  revision 1
 
   bottle do
     revision 1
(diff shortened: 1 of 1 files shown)


mit-scheme.rb
```diff
@@ -1,10 +1,10 @@
-require "formula"
-
 class MitScheme < Formula
-  homepage "http://www.gnu.org/software/mit-scheme/"
-  url "http://ftpmirror.gnu.org/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
-  mirror "http://ftp.gnu.org/gnu/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
-  sha1 "d2820ee76da109d370535fec6e19910a673aa7ee"
+  desc "MIT/GNU Scheme development tools and runtime library"
+  homepage "https://www.gnu.org/software/mit-scheme/"
+  revision 1
```
DomT4: `revision` should always be the last line of the initial block.

mit-scheme.rb
```diff
@@ -1,10 +1,10 @@
...
+  url "https://ftpmirror.gnu.org/mit-scheme/stable.pkg/9.2/mit-scheme-c-9.2.tar.gz"
```
DomT4: This should remain `http` - The `ftpmirror` doesn't support SSL/TLS yet, sadly.

DomT4:
> Don't patch src/lib/include/configure

Are there any runtime consequences to this?

L2G:
I didn't notice any runtime consequences, and I run Homebrew outside of `/usr/local` whenever I'm doing development like this.

mit-scheme.rb
L2G: Should we maybe make the ftp.gnu.org URL the primary one, then?

DomT4: GNU fairly strongly push people onto using the geolocating one and only using the core GNU one as a fallback. We've tried to respect that until GNU indicate otherwise.

L2G:
I run Homebrew outside of `/usr/local` all the time; I have a development environment set up for it and all its taps that lives outside of `/usr/local`.

I dropped the patching of src/lib/include/configure because of issue #24. Apart from `ac_default_prefix`, which is overridden when the script is run with `--prefix`, the only other places where `/usr/local` is present is in some hard-coded paths to search for X headers, and I don't think there's any point in changing those anyway.

microcode/configure maybe shouldn't be patched either, but I haven't investigated that since "it ain't broke."

mit-scheme.rb
L2G: I was counting on `brew audit --strict` to catch gaffes like this. Or is there something else that's checking the order of these lines?

DomT4: It should be. Perhaps we aren't picking up `revision` placement for some reason.

L2G:
While we're here…

Does anyone know the nature of the X11 dependency? I've been able to build it just fine without XQuartz on my system.

DomT4:
> Does anyone know the nature of the X11 dependency?

No idea. Does anything change about the end product if you install with/without `x11`?

L2G:
@DomT4: I'm trying a couple of new builds to investigate this.

I saw in the release notes for MIT-Scheme 9.0 that the X11 dependency is now run-time rather than build-time, so I'm not expecting to find a difference.

DomT4:
Thanks @L2G. If we fix the nits we discussed further up I'm happy to merge this in.

If they're still declaring an X11 dependency, just at runtime rather than buildtime, I guess we keep the dependency since strictly the software will fail without X11 on the system :octopus:.

L2G:
Sorry, I wasn't clear. It is intended to work with or without X11. It will not break in the absence of X11.

DomT4:
Even at runtime? You can remove X11 from the system and still use all the executables/etc shipped?

L2G:
They say the X11 support is "dynamically loadable" now; that seems a better way to describe it than as a "dependency."

See the [release notes](https://www.gnu.org/software/mit-scheme/release.html) for 9.1, under "Minor changes and bug fixes": "X11 support is now dynamically loadable, not a compile-time option."

DomT4:
Sadly we can only describe things as a dependency or not a dependency. I guess perhaps it'd work as `depends_on :x11 => :recommended` or even `:optional`? It's an unusual setup.

@mikemcquaid - If you want to jump in on this discussion re how/what we declare, please feel free.

L2G:
If we have to define any kind of dependency, I think `:optional` would be the most appropriate here.

L2G:
Here's how the executables look from otool's perspective, when built _without_ Xquartz on my system:

```
$ otool -L $(brew ls -v mit-scheme | grep -w bin)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/bchscheme:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/mit-scheme:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/mit-scheme-c:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/scheme:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

After this, I'll post what it looks like _with_ Xquartz. Give me a half an hour or so, since it takes that long to build. :scream:

L2G:
After rebuilding _with_ Xquartz on my system...

```
$ otool -L $(brew ls -v mit-scheme | grep -w bin)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/bchscheme:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/mit-scheme:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/mit-scheme-c:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
/Users/larry/work/GitHub/homebrew/_core_/Cellar/mit-scheme/9.2_1/bin/scheme:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

DomT4:
I'm cool with `:optional` in that case.

Per you & their changelog it seems `mit-scheme` will take advantage of X11 if it is present on the system, so it may be nice to let users know the X11 support hasn't vanished from the project, it's just no longer build-or-run-mandatory. `:optional` is pretty much the easiest way to convey that without going into `caveats` and similar.

L2G:
I finally found the difference (and there is a difference): `lib/mit-scheme-c/lib/prx11.so` gets built and installed when X11 is present; otherwise it does not.

DomT4:
Heh, the subtle differences are always the fun ones to find. Given what upstream said about opportunistic linking rather than during the build I'm wondering if that behaviour isn't in some way a bug.

Let's go `depends_on :x11 => :recommended` for now and keep an eye on the situation going forwards.

mikemcquaid:
Opportunistic linking is a pain. Might be possible to patch the buildsystem to avoid it, though.

DomT4:
Have merged this as-is in https://github.com/Homebrew/homebrew-x11/commit/9506f453158cf28d45d40a537878a075670e8539 since it's a vast improvement over the currently very broken formula.

If someone has the time we should look at the X11 situation and find out whether the creation of `prx11.so` when X11 is found, despite upstream saying they now find X11 opportunistically at runtime and *not* buildtime, is a bug.

PR #85, opened 2015-06-17
yarp: major cleaning up of dependencies and options. No X11 dependency anymore

kt10aan:
Hi,
I have updated the yarp formula to the next version 2.3.64.
This version brought some major changes, the most important of it being the switch from GTK to Qt5.
As a result, X11 is not a dependency anymore and the formula should not be in homebrew-x11.

I would like to ask, after the PR is merged, if the formula can be moved to homebrew-science (yarp is a robotics middleware that is used mainly in research robots in academia).

For upstream details, see https://github.com/robotology/yarp/issues/510

Many thanks :)


kt10aan:
The build failure above is due to an optional dependency on opencv, which is in homebrew/science.
Is this a problem for the PR to be merged? In YARP's documentation, the need for tapping homebrew/science is mentioned. Also, the idea is for the YARP formula to be migrated to homebrew/science anyway.

dunn:
I think changing that line to `depends_on "homebrew/science/opencv" => :optional` would work, at least until it's moved to the science repository.

kt10aan:
@dunn, thanks, I committed the change. Let's hope it passes the test now

yarp.rb
```diff
@@ -1,57 +1,76 @@
...
 
     system "cmake", *args
     system "make", "install"
   end
 
+  def caveats
+    <<-EOS.undent
+      You need to add in your ~/.bash_profile:
+
+          export $YARP_DATA_DIRS=/usr/local/share/yarp
```
DomT4: This won't always be true. You can put Homebrew anywhere, we just have a strong preference for `/usr/local` and it is the default. It'll need to be:

```ruby
def caveats; <<-EOS.undent
  You need to add in your ~/.bash_profile or similar:
    export $YARP_DATA_DIRS=#{HOMEBREW_PREFIX}/share/yarp
  EOS
end
```

kt10aan:
Updated caveat information

yarp.rb
```diff
@@ -1,57 +1,76 @@
...
-      if build.without? "shared"
-        args << "-DCREATE_SHARED_LIBRARY:BOOL=FALSE"
-      else
-        args << "-DCREATE_SHARED_LIBRARY:BOOL=TRUE"
-      end
-      args << "."
+    ]
+
+    if build.with? "qt5"
+      args << "-DCREATE_GUIS=TRUE"
```
DomT4: If there's just one `args` for the block can you turn the block into a line instead? i.e.

```ruby
args << "-DCREATE_GUIS=TRUE" if build.with? "qt5" 
args << "-DENABLE_YARPRUN_LOG=ON" if build.with? "yarprun-log"
args << "-DENABLE_yarpmod_opencv_grabber=ON" if build.with? "opencv"

if build.with? "lua"
  args << "-DYARP_COMPILE_BINDINGS=ON"
  args << "-DCREATE_LUA=ON"
end
```

And such.

yarp.rb
```diff
@@ -1,57 +1,76 @@
...
 
     system "cmake", *args
     system "make", "install"
   end
 
+  def caveats
+    <<-EOS.undent
+      You need to add in your ~/.bash_profile or similar:
+
+          export $YARP_DATA_DIRS=#{HOMEBREW_PREFIX}/share/yarp
```
DomT4: Indentation looks too much here. Should be 2 spaces after the line above starts.

DomT4:
Couple of nits but looking good :+1:. Can you squash your commits when you fix the nits? Thanks!

If @Homebrew/science want the formula now the X11 dep is gone I have no issues facilitating that.

kt10aan:
Fixed the indentations (however, brew audit --strict was not reporting anything wrong before I changed these indentations).
Commits were squashed.
@DomT4, taking care of the move to homebrew/science would be great, thanks :)

kt10aan:
Also, changed the if blocks with one-line args as you suggested

yarp.rb
```diff
@@ -1,57 +1,68 @@
...
 
     system "cmake", *args
     system "make", "install"
   end
 
+  def caveats
+    <<-EOS.undent
+    You need to add in your ~/.bash_profile or similar:
+
+      export $YARP_DATA_DIRS=#{HOMEBREW_PREFIX}/share/yarp
```
dunn: Should the `$` be removed here?

DomT4: Nicely spotted. Yeah, it should be.

DomT4:
> however, brew audit --strict was not reporting anything wrong before I changed these indentations

I believe the `audit` is "trained" to ignore indentation inside an `EOS` block since there's often valid reasons to indent in a particular way (Writing C files for tests, etc).

kt10aan:
Fixed the sign before {HOMEBREW_PREFIX}. It seemed wrong to me too, but it was suggested to me above and I thought it was some Homebrew guideline or something

dunn:
Oh no, the `#` is correct; what's wrong is the `$` in front of `YARP_DATA_DIRS`.

kt10aan:
Ah, yes. Good catch! Fixed.

dunn:
Test failure was just a bot hiccup:
```
 > git checkout -f 27e526a59240c8625029f247569b75412fd2e1ba
FATAL: Could not checkout null with start point 27e526a59240c8625029f247569b75412fd2e1ba
hudson.plugins.git.GitException: Could not checkout null with start point 27e526a59240c8625029f247569b75412fd2e1ba
```

DomT4:
The bot is being sulky this morning. I've re-queued it.

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/0c88bb627d89f01c94e9770fd04eb6267ff8c491. Thanks for your contribution to Homebrew/x11 - We appreciate it :smiley_cat:.

To get discussion in Science rolling you may wish to tee up a PR over there with this formula submitted as a new formula - If they're :+1: on accepting it I'll remove it in `x11` and tweak `tap_migrations.rb` in the core.

kt10aan:
@DomT4, thanks I will create a new PR in Science.

kt10aan:
Opened the PR for the migration: https://github.com/Homebrew/homebrew-science/pull/2427

PR #86, opened 2015-06-18
meld 3.12.3

tschoonj:
version bump
gtk+3 based now
audit --strict compliance

### DIFF
--- a/meld.rb
+++ b/meld.rb
@@ -1,20 +1,29 @@
-require 'formula'
-
 class Meld < Formula
-  homepage 'http://meldmerge.org'
-  url 'http://ftp.gnome.org/pub/GNOME/sources/meld/1.8/meld-1.8.6.tar.xz'
-  sha256 'af96682b8f4bf3ad4221c853b1516218d62a17ff43c38f4a83e7e8ac6736e8a5'
+  desc "A visual diff tool for developers"
+  homepage "http://meldmerge.org"
+  url "https://download.gnome.org/sources/meld/3.12/meld-3.12.3.tar.xz"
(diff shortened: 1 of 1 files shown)


tschoonj:
@DomT4 Any idea whats going on here? Is the PYTHONPATH variable not being set in the test block though I ask for it? The `gi` module should be available...

tschoonj:
Ok never mind, figured out the problem eventually...

DomT4:
> Is the PYTHONPATH variable not being set in the test block though I ask for it? 

As long as you do something like:

```ruby
test do
  ENV["PYTHONPATH"] = "example"
  system "example", "example"
end
```

Then it should be set.

meld.rb
```diff
@@ -1,20 +1,29 @@
...
-  depends_on 'pygobject'
+  depends_on "gtksourceview3"
+  depends_on "pygobject3"
+  depends_on "gobject-introspection"
 
   def install
-    system "make", "prefix=#{prefix}", "install"
-    bin.env_script_all_files(libexec+'bin', :PYTHONPATH => ENV['PYTHONPATH'])
+    ENV.append_path "PYTHONPATH", "#{Formula["libxml2"].opt_lib}/python2.7/site-packages"
+    system "python", "setup.py", "--no-update-icon-cache",
```
DomT4: Will this accept something like `system "python", "--no-update-icon-cache", "--no-compile-schemas", *Language::Python.setup_install_args(prefix)"`?

tschoonj:
No that doesn't work: the icon-cache and schemas arguments have to come between setup.py and install, while prefix has to be last.

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/ca2a8efacf02f15b112f999e10666a55b6fefd62. Thanks @tschoonj!

PR #91, opened 2015-06-23
easy-tag: gnome-icon-theme dependency added

### LINKED ISSUE #89: easy-tag: missing icons after version bump
Most likely [this commit](https://github.com/Homebrew/homebrew-x11/commit/2004b4ce8f712bd3d753553b09907a1c50473a67) is the culprit and the major GTK version bump is to blame.

A picture is worth a thousand words so here goes:
screen shot 2015-06-19 at 09 47 00

True that the version up until now had a couple of icons missing (in the input boxes to apply a change to multiple files) but IIRC most, if not all, the toolbar icons were here.

On a side note, was the upgrade to 2.3.7 necessary? 2.3.x is the unstable branch, the stable one is 2.2.x.

@tschoonj any thoughts?

tschoonj:
Closes #89 
See also discussion in Homebrew/homebrew#40888
Through this dependency, icon support is added to the easy-tag gui as it
has been removed from gtk+3


DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/c32dee649ea47de5b958efbe706546cb90fcecc1. Thanks @tschoonj!

PR #95, opened 2015-06-24
gnumeric: gnome-icon-theme dependency added

tschoonj:
without this dependency, most icons will be missing from the interface
audit --strict compliance


DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/f3059602c292b4c7a0558d11729d994dd2e8ee5f. Thanks @tschoonj!

PR #96, opened 2015-06-25
pdf2svg 0.2.3

UniqMartin:
This release incorporates the changes from the embedded patch and drops the unused GTK+ dependency.

*Side note:* Since the formula no longer depends on GTK+ (which was recently freed of its X11 dependency anyway), would it make sense to move this formula back to the main repository?


UniqMartin:
The bot fails very early on Mountain Lion and Mavericks while attempting to check out the merge commit of the PR. I don't think this is related to my changes and I'm afraid I cannot do anything about this.

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/fd05dcafdbdb4f9651835bb66cf0d3e9d6ac8315. Thank you for your contribution to Homebrew/x11, We appreciate it! :smiley_cat:.

> Since the formula no longer depends on GTK+ (which was recently freed of its X11 dependency anyway), would it make sense to move this formula back to the main repository?

I don't have any huge objection since it doesn't even have GTK linkage now. Want @mikemcquaid to sign off on the idea though.

mikemcquaid:
> Side note: Since the formula no longer depends on GTK+ (which was recently freed of its X11 dependency anyway), would it make sense to move this formula back to the main repository?

:+1: from me

UniqMartin:
Should I leave this to one of the Homebrew maintainers or do it myself? In the latter case, is there some documented procedure for this? If not, my plan would consist of two pieces:
- Homebrew/homebrew: PR that reintroduces the formula and adjusts `tap_migrations.rb`.
- Homebrew/homebrew-x11: PR that removes the formula.

mikemcquaid:
> Homebrew/homebrew: PR that reintroduces the formula and adjusts tap_migrations.rb.
> Homebrew/homebrew-x11: PR that removes the formula.

Yup, that's perfect and it'd be amazing if you could do it :sparkles:

UniqMartin:
Done. See Homebrew/homebrew#41065 and Homebrew/homebrew-x11#97.

Thanks for the guidance and the awesome response times! :+1:

PR #97, opened 2015-06-25
pdf2svg: migrate back to core

UniqMartin:
Turns out `pdf2svg` no longer depends on GTK+ or X11 (and previously did this only by mistake).

See also the related PR Homebrew/homebrew#41065.


tschoonj:
Thanks for the migration. I estimate that currently about one-third of the formulas in homebrew-x11 doesn't belong there given that neither `gtk+` nor `gtk+3` depends on x11...

UniqMartin:
I agree, that seems quite likely. I'm only migrating this one because that's the only one I use from homebrew-x11 and because I was encouraged to do so in Homebrew/homebrew-x11#96. :smiley_cat:

xu-cheng:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

PR #98, opened 2015-06-28
openmotif 2.3.4 (new formula)

DomT4:


tschoonj:
How about submitting this formula to homebrew core instead? With this formula I could add GUI support to `vim`...

DomT4:
The hard X11 dep makes that against our rules for new formulae. Theoretically X11 is optional, but this formula refuses to build without it complaining of missing headers and libs, even with the `--without-x` option passed.

Even if merged in X11 here you can still add it as an option to a core formula - We permit cross-tap `:optional` dependencies.

tschoonj:
good to know I could still use it from within homebrew-core then.

still, I cannot help feeling that a general purpose library like openmotif belongs in homebrew-core, regardless of its x11 dependency...

DomT4:
To be completely honest, if the other popular `motif` solution wasn't development-forever-abandoned status I probably wouldn't merge this formula at this point in time - The need for two patches & removing three test artefacts would be a show-stopper for a lot of new formulae.

I'll ask Mike to comment on the X11/Core situation.

mikemcquaid:
Yeh, I'm probably :-1: on it given the hard X11 dep too.

tschoonj:
Actually you can probably just boneyard lesstif after including this and update the formulas with that dependency. Lesstif was never as good as openmotif and was only started because openmotif was not really open source until recently.

PR #102, opened 2015-07-13
inkscape: remove to core

dunn:
Since it no longer has an X11 dependency.


dunn:
https://github.com/Homebrew/homebrew/pull/41646

dunn:
At some point we'll move this to homebrew/gui, rather than core.

PR #104, opened 2015-07-17
Migrate lesstif to openmotif

mistydemeo:


mistydemeo:
@BrewTestBot test this please

PR #106, opened 2015-07-18
xastir: move to x11, use openmotif

mistydemeo:


mistydemeo:
@BrewTestBot test this please

PR #109, opened 2015-07-19
yarp: add python bindings

kt10aan:
Added ''--with-python'' option.
Python bindings were broken in YARP. Now they are fixed and the option is added in the formula.

### DIFF
--- a/yarp.rb
+++ b/yarp.rb
@@ -3,7 +3,6 @@ class Yarp < Formula
   homepage "http://yarp.it"
   url "https://github.com/robotology/yarp/archive/v2.3.64.tar.gz"
   sha256 "3882b38c39ec9c5fdd06c68f3a4ad21da718bd2733ed7d6a5fb78d9f36dad6b2"
-  head "https://github.com/robotology/yarp.git"
 
   bottle do
     revision 1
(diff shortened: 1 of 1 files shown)


yarp.rb
```diff
@@ -29,7 +30,9 @@ class Yarp < Formula
   depends_on "homebrew/science/opencv" => :optional
   depends_on "qt5" => :optional
   depends_on "lua" => :optional
+  depends_on "python" => :optional
   depends_on "swig" if build.with? "lua"
```
DomT4: Let's block this:

```ruby
if build.with?("lua") || build.with?("python")
  depends_on "swig"
end
```

kt10aan:
Updated :)

DomT4:
Some whitespacing is present apparently:

```
Error: 2 problems in 1 formula
homebrew/x11/yarp:
 * 34: Trailing whitespace was found
 * 61: Trailing whitespace was found
```

kt10aan:
Ooops! Fixed :)

kt10aan:
Regarding the above fail: The ``--with-python`` option works only with the ``--HEAD`` option.
What is the recommended way to add the option only if the formula is being built from ``--HEAD``

DomT4:
Tuck everything inside a `head do` block. You'll need to revert the `swig` change as well. Something like:

```ruby
head do
  url "blah"

  option "with-python", "Install Python Bindings"

  depends_on "python" => :optional
  depends_on "swig" if build.with? "python"
end
```

And then in `def install`:

```ruby
if build.head? && build.with?("python")
  blah
  blah
end
```

Should work fine.

kt10aan:
Thanks for the help! I updated the formula according to your suggestions. The checks failed but I think it is the browbot acting up.

DomT4:
@BrewTestBot test this please

bfontaine:

dunn:
Merged in e5698cd8942b8f47440dec86107b920e357046d0, thanks!

Could you ask upstream to tweak their linking before the next stable release?
```
Warning: python modules have explicit framework links
These python extension modules were linked directly to a Python
framework binary. They should be linked with -undefined dynamic_lookup
instead of -lpython or -framework Python.
  /usr/local/Cellar/yarp/HEAD/lib/python2.7/site-packages/_yarp.so
```

PR #127, opened 2015-08-30
bochs: Added support for customizing debugger

ibalajiarun:
--with-gdb-stub will enable GDB debugger
--with-debugger will enable generic debugger

### DIFF
--- a/bochs.rb
+++ b/bochs.rb
@@ -21,6 +21,7 @@ class Bochs < Formula
     args = %W[
       --prefix=#{prefix}
       --with-x11
+      --with-nogui
       --enable-disasm
       --disable-docbook
       --enable-x86-64
--- a/bochs.rb
+++ b/bochs.rb
@@ -44,7 +44,7 @@ class Bochs < Formula
       --enable-sb16
       --enable-clgd54xx
       --with-term
-      --enable-ne2000"
+      --enable-ne2000
     ]
     
     if build.with? "gdb-stub"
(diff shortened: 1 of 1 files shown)


bochs.rb
```diff
@@ -13,11 +13,20 @@ class Bochs < Formula
...
+    opts << "--enable-gdb-stub"
+  else
+    opts << "--enable-debugger"
+  end
 
   def install
     system "./configure", "--prefix=#{prefix}",
                           "--with-x11",
-                          "--enable-debugger",
+                          *opts,
```
dunn: Ruby 1.8 on Mountain Lion can't deal with a splat in the middle of a list, it'll need to be moved to the end.

bochs.rb
```diff
@@ -13,11 +13,20 @@ class Bochs < Formula
   depends_on "pkg-config" => :build
   depends_on :x11
   depends_on "gtk+"
+  
+  option "with-gdb-stub", "Enable GDB Stub"
+  
+  opts = []
```
DomT4: This is in the wrong place, you need to pass arguments inside `def install`, not outside of it. Let's also not create an empty array just for one option.

Do something like this:

```ruby
def install
  args = %W[
    --prefix=#{prefix}
    --with-x11
    --enable-disasm
    --disable-docbook
    --etc
    --etc
    --etc
  ]

  if build.with? "gdb-stub"
    args << "--enable-gdb-stub"
  else
    args << "--enable-debugger"
  end
```

bfontaine:

apjanke:

apjanke:

PR #128, opened 2015-09-12
hatari: migrate to homebrew-games

tomyun:
Since Hatari builds and runs fine without x11 dependency, I'd suggest migrating it to `homebrew-games`. Homebrew/homebrew-games#380 is a PR for new formula that contains some updates including version bump to 1.9.0.


DomT4:
Are you happy to take this in Games @mistydemeo?

PR #129, opened 2015-09-21
zenity 3.18.0

tschoonj:


DomT4:
One of the dependencies is having build issues:

```
checking for xml2-config... /usr/bin/xml2-config
checking libxml2 version... 2.9.2
checking for xslt-config... /usr/bin/xslt-config
checking which XML catalog to use... /usr/local/etc/xml/catalog
checking for DocBook XML DTD... configure: error: not found. Make sure you have the DocBook DTD installed and ensure that it is registered in /usr/local/etc/xml/catalog.
==> Formula
Path: /usr/local/Library/Formula/scrollkeeper.rb
```

tschoonj:
Interesting

I am still running Yosemite so I cannot test this unfortunately :-(
Any chance you could help out here?

DomT4:
This didn't reproduce for me on a system which had a prior `/usr/local/etc/xml/catalog` on Yosemite before upgrading, but if I remove that and then reinstall `docbook` & `docbook-xml` I hit the exact error. So looks like it's failing to do something on El Cap, although quite what isn't obvious.

tschoonj:
I think that the `docbook` and `docbook-xsl` scripts should work on `HOMEBREW_PREFIX/etc/xml/catalog` instead of `#{etc}/xml/catalog`

Ideally, when uninstalling/unlinking `docbook` or `docbook-xsl` there would be a `post_uninstall` method that gets called to update `HOMEBREW_PREFIX/etc/xml/catalog` accordingly

DomT4:
`#{etc}` = `HOMEBREW_PREFIX/etc`

tschoonj:
Didn't know that :-)

But shouldn't `docbook`'s catalog manipulations happen in a `post_install` block then? Otherwise you will go out of the sandbox right?

DomT4:
@BrewTestBot test this please

tschoonj:
@DomT4 this one also looks ready :smile:

DomT4:
Indeed :wink:. Merged in https://github.com/Homebrew/homebrew-x11/commit/cdd4d03f1b1dfb282230548a3474a7a0c9c48f90. Thanks again Tom!

PR #132, opened 2015-09-27
Updated to 15.08 release.

manuseiffel:


dunn:
Merged in f10c20887f59a9ab440c2d4c1b7d3d129f6639ee, thank you!

dmzimmerman:
Unfortunately this version, like previous versions, does not actually work when built - it triggers assertion failures in gdkimage-quartz.c and the user interface is rendered unusable. See the discussion on https://github.com/Homebrew/homebrew-x11/pull/83

dunn:
Do the command-line tools work?

dmzimmerman:
They appear to; I can definitely build example projects without any issues. However, I was intending to use EiffelStudio on my current project mainly for the visual modeling features, so the user interface was the most important part for me.

manuseiffel:
How do I do to enable the quartz back end to see how we could work around the bug in this back end?

Manu

dmzimmerman:
I'm not sure... I have little experience with building gtk+, I just found the other issue by searching. I think the configure script parameters are in the gtk+.rb file in the main Homebrew repo, and that might be a place to start.

dunn:
It this hasn't been reported upstream, it'd be great if one of you could do that as well.

manuseiffel:
It will take me a few weeks to be able to test this unfortunately. So stay tuned.

 

Manu

PR #133, opened 2015-10-02
libsigc++ dependent formulas need C++11 support

tschoonj:
Recently all C++ gnome packages introduced a mandatory C++11 requirement,
which implies that all packages depending on these also need to have
the same requirement.

This PR add revision bumps as well as makes the necessary changes for
C++ compiler support


DomT4:
Merged. Try to keep the commit message consistent with our usual format if you can; something like:

```
formula-xyz: revision for mandatory c++11 support
```
With one commit per formula would have been cool here.

tschoonj:
Thanks for merging it in. Sorry about the commit message. Will keep it mind for next time

PR #134, opened 2015-10-03
terminator: 0.98, add HEAD

apjanke:


dunn:
Merged in 9f0fcfa4f22e19d9267dde02ef2a0c60e15a0337, thanks! No need to knock out the bottle revision, that's also handled by the bot.

PR #138, opened 2015-10-08
freerdp: depend on openssl for MacOS > yosemite

bmiklautz:
FreeRDP requires openssl. For MacOS.version <= yosemite keep using the
systems version as before. For El Capitan and on use the dependency on
openssl.

### DIFF
--- a/freerdp.rb
+++ b/freerdp.rb
@@ -1,5 +1,6 @@
 class Freerdp < Formula
   homepage "http://www.freerdp.com/"
+  revision 1
 
   stable do
     url "https://github.com/FreeRDP/FreeRDP/archive/1.0.2.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Let's just make it unconditional. We don't support using the deprecated, archaic system OpenSSL.

bmiklautz:
@DomT4 fair enough :). I've updated the pull request.

DomT4:
Can you add a `revision 1` line as well please?

```diff
 homepage "http://www.freerdp.com/"
+revision 1

stable do
  url "https://github.com/FreeRDP/FreeRDP/archive/1.0.2.tar.gz"
```

Means everyone is moved off the system OpenSSL at the same time.

dunn:
Thanks!! Merged in 1e145d910657f80d5e2c5a02db3433b155a1f941.

PR #139, opened 2015-10-12
Add universal option to openmotif (requires universal build of flex)

dgarnier:


openmotif.rb
```diff
@@ -14,11 +14,14 @@ class Openmotif < Formula
   depends_on "autoconf" => :build
   depends_on "libtool" => :build
   depends_on "pkg-config" => :build
+  depends_on "flex" => :build
```
dunn: If it needs to be universal, but only when the formula is build `--universal`, this should be
```ruby
if build.universal?
  depends_on "flex" => [:build, :universal]
end
```

dgarnier: So true.  Fixed.. Thanks.

dunn:
Thanks so much! Merged in dc9636a965f3e883fa6c84b160028110fc750a44.

PR #140, opened 2015-10-13
swi-prolog 7.3.9 (devel) and 7.2.3 (stable)

honnix:
Upgrade swi-prolog to development version 7.3.9 and stable version 7.2.3


dunn:
Feel like adding a better test while you're at it :smile_cat: ??

honnix:
Thanks for the suggestion.

Yes, that I can do. But not exactly sure what could be better. Because as long as "swipl --version" can execute successfully, the installation should be fine. There is only one dylib depended by swipl binary.

Is there any general guide from Homebrew regarding how/what to test against?

dunn:
We usually try to perform some basic real-life operation; for libraries, we compile a minimal program that uses it; for applications we try to have it perform some common function(s) https://github.com/Homebrew/homebrew/blob/master/Library/Formula/scons.rb#L34-L45

honnix:
OK. I will give it a try. 



Best regards,
Hongxin

honnix:
I added one test case by feeding in a prolog file and asking it to solve a simple goal. How can I trigger the test process?

DomT4:
Merged in https://github.com/Homebrew/homebrew-x11/commit/b271cc2c2112909c9ec4196a4c1ab2e41b56753a. Thank you for your contribution to Homebrew @honnix! :smiley_cat:

PR #141, opened 2015-10-15
x3270 3.4ga8

josa42:


x3270.rb
```diff
@@ -17,20 +17,20 @@ class X3270 < Formula
...
   end
 
-  def install
-    make_directory "x3270-3.3"
-    make_directory "c3270-3.3" if build.with? "c3270"
-    make_directory "pr3287-3.3" if build.with? "pr3287"
-    make_directory "s3270-3.3" if build.with? "s3270"
-    make_directory "tcl3270-3.3" if build.with? "tcl3270"
+  test do
+    system bin/"x3270", "--version"
```
dunn: This is better than nothing, but would it be possible for something that checks a bit more of the functionality? Thanks!

josa42: Yeah I know it is not ideal. But I could come up with something better.

dunn:
Thanks for cleaning up this formula! Merged in 141bb90a3e57bef7f6da83a23076bb5ac106161d.

PR #142, opened 2015-10-16
euslisp 9.15 (new formula)

wkentaro:


wkentaro:
How can I avoid this error?
```
 * Non-libraries were installed to "/usr/local/Cellar/euslisp/9.15/lib"
Installing non-libraries to "lib" is discouraged.
The offending files are:
  /usr/local/Cellar/euslisp/9.15/lib/eus.init.l
  /usr/local/Cellar/euslisp/9.15/lib/eusglrt.l
  /usr/local/Cellar/euslisp/9.15/lib/EUSRC
  /usr/local/Cellar/euslisp/9.15/lib/eusrt.l
```

dunn:
Are those manpages? One way would be to `man1.install` each of them after the make step; but better would be to ask the developer to tweak their Makefile.

wkentaro:
They are lisp files, and also there are other directories which contains lisp files. (https://github.com/euslisp/EusLisp/tree/master/lib)
so I think changing it is difficult from historical reason.

dunn:
Oh, those are `l`s, not `1`s.

I'm not sure what the best approach is here.  Could you ask upstream how difficult it would be to install them elsewhere?

wkentaro:
> Oh, those are ls, not 1s.

Actually.. :smiley: 

> Could you ask upstream how difficult it would be to install them elsewhere?

Sure.

wkentaro:
I got comment [here](https://github.com/euslisp/EusLisp/issues/82#issuecomment-148978578) from one of the maintainers of euslisp.
So now I'm going to change the directory construction to avoid this error.

dunn:
Thanks! Just the smallest of requests:
```
== /usr/local/Library/Taps/homebrew/homebrew-x11/euslisp.rb ==
C: 15: 28: Unnecessary spacing detected.

1 file inspected, 1 offense detected

**Comment by wkentaro** [comment]
I fixed it.

File: euslisp.rb [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+  depends_on :x11
+
+  def install
+    ENV.deparallelize
+    ENV.O0
+
+    cd "#{prefix}" do
+      cp_r buildpath, "eus" # euslisp needs to be compiled in Cellar
+
+      cd "eus" do
```
dunn: It might be possible to combine these two `cd`s into
```ruby
mkdir prefix/"eus" do
  cp_r buildpath, prefix/"eus"

wkentaro:
I updated.

dunn:
Thank you! Merged in 4143ee1cabdb6f4e818cc024cd3f6fb88b1f75c0.

PR #143, opened 2015-10-16
jskeus 1.0.10 (new formula)

wkentaro:

PR #150, opened 2015-11-08
euslisp 9.15: create wrapper script that sets env

wkentaro:


wkentaro:
Will be merged in #143

PR #172, opened 2016-01-02
xlispstat 3-52-23 (new formula)

jhbadger:
This is a formula for xlispstat, a data-analysis and plotting platform similar to R but based on Lisp.

### DIFF
--- a/bochs.rb
+++ b/bochs.rb
@@ -1,6 +1,6 @@
 class Bochs < Formula
+  desc "Open source IA-32 (x86) PC emulator written in C++"
   homepage "http://bochs.sourceforge.net/"
-  desc "A highly portable open source IA-32 (x86) PC emulator written in C++"
   url "https://downloads.sourceforge.net/project/bochs/bochs/2.6.8/bochs-2.6.8.tar.gz"
   sha256 "79700ef0914a0973f62d9908ff700ef7def62d4a28ed5de418ef61f3576585ce"
 
@@ -11,14 +11,13 @@ class Bochs < Formula
     sha256 "93613f9e68caedc20b4ae3ea3b5ea7b1e5b4450e8c1319cdd3a5e364111fd67c" => :mavericks
   end
 
+  option "with-gdb-stub", "Enable GDB Stub"
+
   depends_on "pkg-config" => :build
   depends_on :x11
   depends_on "gtk+"
 
-  option "with-gdb-stub", "Enable GDB Stub"
-
   def install
-
     args = %W[
       --prefix=#{prefix}
       --with-x11


jhbadger:
Can someone explain what "brew uses --recursive" is doing and why it is failing in the above test? I understand that the first commit didn't like the style of my test and so I modified it, but I have no idea what is going wrong in "brew uses --recursive". I have no problems installing xlispstat on my own systems using my formula.

DomT4:
@BrewTestBot test this please

DomT4:
Apologies, the `--recursive` error was caused by a circular dependency in this tap we hadn't picked up on till it started killing builds. Resolved via https://github.com/Homebrew/homebrew-x11/commit/ef84174ccef5394416eccd2869130956b48c9409.

xlispstat.rb
```diff
@@ -0,0 +1,19 @@
...
+  desc "Statistical data science environment based on Lisp"
+  homepage "http://homepage.stat.uiowa.edu/~luke/xls/xlsinfo/"
+  url "http://homepage.cs.uiowa.edu/~luke/xls/xlispstat/current/xlispstat-3-52-23.tar.gz"
+  version "3-52-23"
+  sha256 "9bf165eb3f92384373dab34f9a56ec8455ff9e2bf7dff6485e807767e6ce6cf4"
+  depends_on :x11
+
+  def install
+    system "./configure", "--prefix=#{prefix}", "--disable-debug", "--disable-dependency-tracking"
+    ENV.j1 # Or make fails bytecompiling lisp code
```
dunn: Should this be considered a bug? If so it'd be great to have it reported upstream.

bfontaine:
Ping re Alex’ comment above?

jhbadger:
1) I'm not so sure it can really be considered a "bug" given that it only happens with the optional arguments to make that homebrew uses rather than with the method supported by the author

2) Xlispstat isn't really under much (if any) current development -- the most recent version is from 2013 and before that, 2003. It was a popular statistical system in the 1990s before the advent of R but R has really taken most of the statistical field these days. I still think it has value as its dynamic graphs are something that hasn't really been equalled in R, plus older statistics books often give examples in xlispstat rather than R, but I understand if homebrew doesn't want to handle such legacy software.

Jonathan

apjanke:
@BrewTestBot test this please

xlispstat.rb
```diff
@@ -0,0 +1,19 @@
+class Xlispstat < Formula
+  desc "Statistical data science environment based on Lisp"
+  homepage "http://homepage.stat.uiowa.edu/~luke/xls/xlsinfo/"
+  url "http://homepage.cs.uiowa.edu/~luke/xls/xlispstat/current/xlispstat-3-52-23.tar.gz"
+  version "3-52-23"
```
dunn: Looks like the version should be `3.52.23`: http://homepage.cs.uiowa.edu/~luke/xls/xlispstat/current/RELEASE-3-52

xlispstat.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def install
+    system "./configure", "--prefix=#{prefix}", "--disable-debug", "--disable-dependency-tracking"
+    ENV.j1 # Or make fails bytecompiling lisp code
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "echo", "\"(median (iseq 1 100))\" | #{bin}/xlispstat"
```
dunn: You can use `pipe_output` instead of echo, and check the result with

jhbadger:
What do these errors mean? I'm not sure how to interpret them. The formula (with the assert & version number changes suggested above) works on my system.

DomT4:
You've pulled so much into this PR (I'd guess accidentally, looks like a rebase gone bad) that the CI can't pull it cleanly. You can fix it by rebasing against an up-to-date master, usually.

apjanke:
The source for this PR is `jhbadger:master`. Making a PR off the master branch is prone to pulling in extra changes. PRs should be made off a dedicated feature branch. See the [Homebrew Pull Request guide](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-%28and-get-it-merged%29.md).

Easiest way to fix this is probably to make a new branch, cherry-pick the changes for `xlispstat` to it, and create a new PR from the branch.

PR #173, opened 2016-01-05
xephem 3.7.7 (new formula)

liljenstolpe:
XEphem is a x-based astronomy package.

### DIFF
--- a/xephem.rb
+++ b/xephem.rb
@@ -1,0 +1,17 @@
+class Xephem < Formula
+  desc "The Xephem Astronomy Package"
+  homepage "http://www.clearskyinstitute.com/xephem/index.html"
+  url "http://97.74.56.125/free/xephem-3.7.7.tar.gz"
+  sha256 "d1f8e17cfc5d2e3af5fd5a8bcf34bbf99a79d40f66326c098a819f82af62b4b7"
+
+  depends_on :x11
+  depends_on "libpng"
+
+  def install
+    cd "GUI/xephem"
+    system "make", "MOTIF=../../libXm/osx"
+    bin.install "xephem"
+    man.mkpath
+    man1.install "xephem.1"
+  end
+end


liljenstolpe:
Any suggestions for a test_block?  xephem ALWAYS trys and opens the DISPLAY, and the Jenkins environment doesn't seem to have an xserver running.

liljenstolpe:
The audit fails as I don't have a 'test do' block.  However, xephem ALWAYS tries to open $DISPLAY, and errors out when it can't.  Since the jenkins environment doesn't seem to start an xserver, this always fails.

liljenstolpe:
There are two issues here.

1) Since the jenkins environment doesn't start an xserve, I can't do a "test do" block as xephem ALWAYS tries to open the $DISPLAY, even if it's just printing on the terminal - so it always fails.

2) The png code relies on a very old version of the png library.  I can build on my machine, but it looks like it fails with the 'depends_on "libpng"' included version.

I will close until the upstream is fixed, but would like some help with issue (1) in the mean time.

DomT4:
@BrewTestBot test this please

xephem.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 "d1f8e17cfc5d2e3af5fd5a8bcf34bbf99a79d40f66326c098a819f82af62b4b7"
+
+  depends_on :x11
+  depends_on "libpng"
+
+  def install
+    cd "GUI/xephem"
+    system "make", "MOTIF=../../libXm/osx"
+    bin.install "xephem"
+    man.mkpath
```
bfontaine: You can remove this line, `man1.install` below will create the necessary directories.

xephem.rb
```diff
@@ -0,0 +1,17 @@
...
+  desc "The Xephem Astronomy Package"
+  homepage "http://www.clearskyinstitute.com/xephem/index.html"
+  url "http://97.74.56.125/free/xephem-3.7.7.tar.gz"
+  sha256 "d1f8e17cfc5d2e3af5fd5a8bcf34bbf99a79d40f66326c098a819f82af62b4b7"
+
+  depends_on :x11
+  depends_on "libpng"
+
+  def install
+    cd "GUI/xephem"
```
bfontaine: Please use a block:
```rb
cd "GUI/xephem" do
  # ... code ...
end
```

bfontaine:
> 1) Since the jenkins environment doesn't start an xserve, I can't do a "test do" block as xephem ALWAYS tries to open the $DISPLAY, even if it's just printing on the terminal - so it always fails.

In this case we could make a dummy test that checks if e.g. the binary exists: `assert (bin/"...").exist?`.

apjanke:
Closing due to lack of response.

PR #174, opened 2016-01-06
mupdf 1.8

proinsias:
mupdf 1.8 is now available – formula has been updated accordingly.

I've also changed the link to point to the `downloads/archive` folder,
rather than the `downloads` folder – using the latter folder results in dead
links when a new version replaces the old version in the latter
folder.

brew test mupdf fails because the `mudraw` command is missing, even
though there are no build errors.

I'd appreciate help figuring this out – this is my first homebrew
commit/pull request.


proinsias:
Whoops. Didn't see there already was a pull request for this.

PR #176, opened 2016-01-10
X11 migration

tschoonj:
moving all Formulas that depend on gtk+ or gtk+3 from x11 to core

counterpart of Homebrew/homebrew#47924


DomT4:
And merged! Thanks again Tom!

PR #183, opened 2016-02-07
libview 0.6.6 (new formula)

vicamo:
Signed-off-by: You-Sheng Yang (楊有勝) <vicamo@gmail.com>

### DIFF
--- a/libview.rb
+++ b/libview.rb
@@ -1,0 +1,45 @@
+class Libview < Formula
+  desc "VMware's incredibly exciting widgets"
+  homepage "http://view.sourceforge.net/"
+  url "http://httpredir.debian.org/debian/pool/main/libv/libview/libview_0.6.6.orig.tar.gz"
+  sha256 "394583d2d28c334684d512eb2a260db51a877b84319f74a2425b79b5780f3ad8"
+  head "svn://svn.code.sf.net/p/view/code/trunk/libview"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "gtk+"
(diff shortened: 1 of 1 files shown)


tschoonj:
This formula is a bit problematic IMHO.

The project appears to be dead as no releases have been made since May 2010. Also I have doubts as to how it could properly work: it has a hard dependency on `:x11` but `gtk+` does not have the x11 backend anymore so I guess that part of the code would never work (and probably lead to segfaults after going quickly going through the source).

@vicamo Can you verify if this package actually works?

@bfontaine FYI

vicamo:
@tschoonj in this package I also install its demo programs, so you can really have a try yourself.

tschoonj:
The tests do not cover the problematic widgets MotionTracker and ToolTip that use the X11 code.

Do you have a project where you use these? From the code it looks like they will result very easily in a segmentation fault.

vicamo:
@tschoonj https://github.com/Homebrew/homebrew/pull/48814 contains an example tool `gvncviewer` that can be configured to use libview. It's not enabled now because libview is not available yet. And, yes, gvncviewer uses only __AutoDrawer__.

tschoonj:
`gtk-vnc` depends on `gtk+3` while `libview` depends on `gtk+` through `gtkmm`. Since you should never mix gtk+2 and gtk+3 code you cannot build `gtk-vnc` with `libview` support.

I suggest you remove the problematic files from `libview/Makefile.in` and just compile without them.

MikeMcQuaid:
Closing due to @tschoonj's comments.

PR #187, opened 2016-02-23
rxvt-unicode: add unicode3 support

### LINKED ISSUE #167: rxvt-unicode: --enable-unicode3
Added --enable-unicode3 to list of configure options. This was needed for things like vim airline to look right. Please consider adding this option by default?

apjanke:
Build with `--enable-unicode3` by default. Exposes `--without-unicode3` so users
can get a build that matches the rxvt-unicode distribution's default.

Closes #167


apjanke:
Pinging @DomT4 @dunn as prior authors on this formula: does this (`--enable-unicode3` by default, with formula option to disable) sound sensible?

DomT4:
Seems fine to me based off your benchmarking in the other thread :+1:.

PR #191, opened 2016-02-24
tiger-vnc 1.6.0

mpfz0r:
works fine, and 1.4.0 broke with a recent gnutls update.


apjanke:
@BrewTestBot test this please

apjanke:

The build failure happened because fltk is not available on 10.10+. Ignoring it; Mavericks bottle is still available.

PR #195, opened 2016-03-11
swi-prolog 7.3.18 (devel)

honnix:
Besides the version, a fix for compiler optimization is done. Get
details here: SWI-Prolog/packages-semweb#22


swi-prolog.rb
```diff
@@ -62,13 +62,16 @@ def install
     ENV.append "DISABLE_PKGS", "jpl" if build.without? "jpl"
     ENV.append "DISABLE_PKGS", "xpce" if build.without? "xpce"
 
+    # Default value Os causes segfault
+    ENV["HOMEBREW_OPTIMIZATION_LEVEL"] = "O2"
```
DomT4: Should be able to shorten this to `ENV.O2`? I'd be a bit surprised if `superenv` honoured formula overriding of the `HOMEBREW_OPTIMIZATION_LEVEL` variable.

honnix: So change `ENV["HOMEBREW_OPTIMIZATION_LEVEL"] = "O2"` to just `ENV.O2`?

I don't really know how this `ENV` thing works to be honest. This PR is a more like a hack than proper code. I can try it out and resubmit the PR. On the other hand, we are still investigating the root cause because modifying compiler flag in this case is anyway a workaround.

DomT4: I believe just `ENV.O2` is what you're looking for.

honnix: Alright. Thanks a lot! It works.

swi-prolog.rb
```diff
@@ -62,13 +62,17 @@ def install
     ENV.append "DISABLE_PKGS", "jpl" if build.without? "jpl"
     ENV.append "DISABLE_PKGS", "xpce" if build.without? "xpce"
 
+    # Default value Os causes segfault in latest swi-prolog devel version
+    # This is the workaround
+    ENV.O2
```
dunn: Has this been reported upstream?

honnix: Which upstream? Homebrew? Not really. Still not sure it's a compiler problem or
some hidden bug in swi-prolog.

Best regards, Hongxin
On Sat, Mar 12, 2016 at 21:56, Alex Dunn <notifications@github.com> wrote:
In swi-prolog.rb
[https://github.com/Homebrew/homebrew-x11/pull/195#discussion_r55927519] :

> @@ -62,13 +62,17 @@ def install
>      ENV.append "DISABLE_PKGS", "jpl" if build.without? "jpl"
>      ENV.append "DISABLE_PKGS", "xpce" if build.without? "xpce"
>  
> +    # Default value Os causes segfault in latest swi-prolog devel version
> +    # This is the workaround
> +    ENV.O2


Has this been reported upstream?

—
Reply to this email directly or view it on GitHub
[https://github.com/Homebrew/homebrew-x11/pull/195/files#r55927519] .[https://github.com/notifications/beacon/AAJsrBcZhyCsemLDfYT4BZE0exosDU6Rks5psyiQgaJpZM4HufMI.gif]

dunn: Upstream to the swi-prolog developers.

honnix: Sure. I believe they are well aware of this and we are currently trying to
find the root cause.

dunn: Great, thank you!

dunn:
Merged in c14365ed3d3774d701cf5609e87a8b32ae44df90, thanks!

PR #196, opened 2016-03-17
swi-prolog 7.3.19 (devel)

honnix:
Upgrade to latest development version which has fixed
https://github.com/SWI-Prolog/packages-semweb/issues/22

With the fix, workaround playing with optimization flag is no
longer needed.


dunn:
Thank you! Merged as 83e2e9555f8d6f67ffd3bd2fc789dcaf3f453ed1.

Issue #157, opened 2015-11-19
Scantailor, Build error on Mac OS X 10.11.1


davidbecher:
Hey,

I just tried to install scantailor via homebrew but it just won't work. Here some output of the terminal:

```
==> cmake . -DPNG_INCLUDE_DIR=/opt/X11/include -DCMAKE_CXX_FLAGS=-std=c++11 -std
==> make install
Last 15 lines from /Users/dbecher/Library/Logs/Homebrew/scantailor/02.make:
[ 48%] Building CXX object CMakeFiles/stcore.dir/OutputFileNameGenerator.cpp.o
[...]

PR #197, opened 2016-04-08
scantailor EXPERIMENTAL-2016-02-22 (devel)

ilovezfs:
As the upstream commit fixing this says, "With boost 1.60 there's a namespace conflict between boost::bind and boost::lambda::bind and placeholders are no longer in global namespace, so use fully qualified names for these." (Note: This actually affects boost 1.59 as well.)

https://github.com/scantailor/scantailor/pull/194
https://github.com/scantailor/scantailor/commit/2be2b472f0cc35ef82d9e92b2813534aa780b050

New version tag requested: https://github.com/scantailor/scantailor/issues/204

Closes Homebrew/homebrew-x11#157.

### DIFF
--- a/scantailor.rb
+++ b/scantailor.rb
@@ -1,7 +1,11 @@
 class Scantailor < Formula
+  desc "Interactive post-processing tool for scanned pages"
+  homepage "http://scantailor.org/"
+  head "https://github.com/Tulon/scantailor.git"
+
   class Version < ::Version
     def enhanced?
-      to_s.start_with?("enhanced")
+      to_s.start_with?("enhanced", "EXPERIMENTAL")
     end
 
     def <=>(other)
@@ -18,8 +22,6 @@ class Scantailor < Formula
     end
   end
 
-  homepage "http://scantailor.org/"
-
   stable do
     url "https://downloads.sourceforge.net/project/scantailor/scantailor/0.9.11.1/scantailor-0.9.11.1.tar.gz"
     version Scantailor::Version.new("0.9.11.1")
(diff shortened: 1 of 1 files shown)


mikemcquaid:
@ilovezfs Would prefer to see us apply upstream's patch rather than an `inreplace` in these cases. Let's also sit on this and see what happens to the tag request to decide whether to 💀 or not.

ilovezfs:
Unfortunately the upstream patch https://github.com/scantailor/scantailor/commit/2be2b472f0cc35ef82d9e92b2813534aa780b050 doesn't apply cleanly. The change is needed in both stable and devel, but luckily it's always the same three string replaces in the affected files.

The other option would be a vendored boost-1.58 build-time dependency.

ilovezfs:
@mikemcquaid I was pointed to the author's newest work in this comment https://github.com/Homebrew/homebrew-x11/issues/157#issuecomment-207382139

The PR is updated to use the new devel and head versions now, which build cleanly.

scantailor.rb
```diff
@@ -32,6 +34,22 @@ def <=>(other)
...
+    # Approved by maintainer and included in official repository.
+    # See: http://sourceforge.net/p/scantailor/mailman/message/31884956/
+    patch do
+      url "https://gist.githubusercontent.com/muellermartin/8569243/raw/b09215037b346787e0f501ae60966002fd79602e/scantailor-0.9.11.1-clang.patch"
+      sha256 "86a3fa456c146844381af8d5bcdcdb9939edd4b11ce832bd928c1be036a27ea6"
+    end
+
+    # Changes some uses of boost::lambda::bind to C++11 lambdas.
+    # Avoids compilation errors with Boost 1.57.
+    # https://github.com/scantailor/scantailor/issues/125
```
mikemcquaid: Has this been submitted upstream to the project, @jkseppan?

scantailor.rb
```diff
@@ -41,41 +59,55 @@ def <=>(other)
...
+        system "make", "install"
+      end
+    end
   end
 
-  def install
-    system "cmake", ".", "-DPNG_INCLUDE_DIR=#{MacOS::X11.include}", "-DCMAKE_CXX_FLAGS=-std=c++11 -stdlib=libc++", *std_cmake_args
-    system "make", "install"
+  test do
+    spec = Tab.for_name(name).source["spec"]
```
mikemcquaid: Rather than using the tap just use `build.stable?` or `build.devel?` here.

DomT4: Whoops. Apologies for the bad advice yesterday @ilovezfs on using `if stable?` here. `build.stable` will actually work inside the test, but doesn't for `postinstall` as documented [here](https://github.com/Homebrew/legacy-homebrew/issues/37120).

scantailor.rb
```diff
@@ -41,41 +59,55 @@ def <=>(other)
...
   depends_on "libtiff"
   depends_on :x11
-
-  # Makes Scan Tailor work with Clang on OS X Mavericks
-  # Approved by maintainer and included in official repository.
-  # See: http://sourceforge.net/p/scantailor/mailman/message/31884956/
-  patch do
-    url "https://gist.githubusercontent.com/muellermartin/8569243/raw/b09215037b346787e0f501ae60966002fd79602e/scantailor-0.9.11.1-clang.patch"
-    sha256 "86a3fa456c146844381af8d5bcdcdb9939edd4b11ce832bd928c1be036a27ea6"
+  if build.stable?
```
DomT4: These still belong in the `stable do` and `devel do` blocks though :trollface:

ilovezfs:
Author tagged a new stable spec for us, so I'll open a fresh PR with just that.

PR #202, opened 2016-04-23
mit-scheme: avoid -march=native

ilovezfs:
Setting -march=native, which is what --build-from-source does, can fail
with an error that looks like this:

  The object (#[compiled-closure 12 ("rtlcon" #x1) #x7 #xdf2f2 #x68483f]
  #[compiled-procedure 13 ("cfg3" #x23) #x3 #xe29e2] plus-fixnum #[false
  14 #x40202] . #[false 15 #x147a3]), passed as the second argument to
  apply, is not the correct type.

Only Haswell and above (i.e., Haswell, Broadwell, and Skylake currently)
appear to be impacted.

Reported 23rd April 2016 to https://savannah.gnu.org/bugs/index.php?47767

Also, fix "configure: error: No MacOSX SDK for version: 10.10."
Reported 23rd April 2016: https://savannah.gnu.org/bugs/index.php?47769


mikemcquaid:
Looks good and thanks for reporting upstream.

ilovezfs:
Great, and you're welcome :)

PR #205, opened 2016-05-02
scantailor 0.9.12.2

ilovezfs:


ilovezfs:
@DomT4 @apjanke @mikemcquaid This is the PR superseding https://github.com/Homebrew/homebrew-x11/pull/197

apjanke:
:+1:

ilovezfs:
@apjanke I modified the test because the upstream version numbering is unreliable so the simple version test isn't a great plan. Does that pseudo-functional test look all right?

In this case the author retagged 0.9.12.1 with the tag RELEASE_0_9_12_2 because the initial tag didn't bump the so-called "quad version" to 0.9.12.1 in version.h.

CLI reports 0.9.12 (the non-quad version), but 0.9.12 is actually a different and earlier tag. Upstream's view is that this release is "0.9.12.1" but I'm tempted to remove the explicit version number, and let brew think it's "0.9.12.2" since there presumably will never be a "real" 0.9.12.2 given that the RELEASE_0_9_12_2 is already used up.

apjanke:
Yeah, that sounds fine. The new test looks more meaningful even if the versions were reliable. :+1:

PR #208, opened 2016-05-10
mit-scheme: migrate to core

woodruffw:
Removes mit-scheme from the x11 tap, pending Homebrew/homebrew-core#1030.

cc @UniqMartin @ilovezfs


dunn:
Merged as 0293c55e272acfff9b24f93f6a2a9827774479d1, thanks so much!

PR #211, opened 2016-05-23
chezscheme 9.4 (new formula)

VMatrix1900:
Chez Scheme is a compiler and run-time system for the language of the
Revised6 Report on Scheme (R6RS), with numerous extensions. The compiler
generates native code for each target processor, with support for x86,
x86_64, and 32-bit PowerPC architectures.

### DIFF
--- a/chezscheme.rb
+++ b/chezscheme.rb
@@ -1,0 +1,27 @@
+class Chezscheme < Formula
+  desc "Chez Scheme"
+  homepage "https://cisco.github.io/ChezScheme/"
+  url "https://github.com/cisco/ChezScheme/archive/v9.4.tar.gz"
+  sha256 "9f4e6fe737300878c3c9ca6ed09ed97fc2edbf40e4cf37bd61f48a27f5adf952"
+
+  depends_on :x11 => :build
+
+  conflicts_with "mit-scheme", :because => "both install `scheme` binaries"
+
+  def install
+    system "./configure", "--installprefix=#{prefix}"
+    system "make", "install"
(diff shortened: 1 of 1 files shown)


michaelballantyne:
@VMatrix1900: I was just about to submit a PR for this but you beat me to it! How would you feel about adding "--threads" and "--installschemename=chez" to the configure and dropping the conflict with MIT Scheme?

I think it would also probably be better to make the X11 dependency optional (per https://github.com/cisco/ChezScheme/issues/9) and put this in the homebrew-core repository, but that would require some hacking on chez's build system.

michaelballantyne:
Or if we wanted to hack it, the formula could comment out the include mentioned in that ticket (`#define LIBX11 "/usr/X11R6/lib/libX11.dylib"`) in `version.h` if the X11 option is off.

VMatrix1900:
@michaelballantyne : Thanks for your suggestion. Upstream may add a configure option to disable the X11 build dependency(cisco/ChezScheme#9). Let's wait for it.

MikeMcQuaid:
@VMatrix1900 Thoughts on that? Want to get this merged or closed, I think.

VMatrix1900:
@MikeMcQuaid  still waiting for the upstream update. But I think it's okay to merge.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #212, opened 2016-06-07
curlftpfs: fix head url

ilovezfs:
https://github.com/rfw/curlftpfs.git no longer exists

### DIFF
--- a/curlftpfs.rb
+++ b/curlftpfs.rb
@@ -3,7 +3,7 @@ class Curlftpfs < Formula
   url "https://downloads.sourceforge.net/project/curlftpfs/curlftpfs/0.9.2/curlftpfs-0.9.2.tar.gz"
   sha256 "4eb44739c7078ba0edde177bdd266c4cfb7c621075f47f64c85a06b12b3c6958"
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 any idea if we have a way to tell CVS to use the password blank?

DomT4:
We can do some [automated](https://github.com/Homebrew/homebrew-core/blob/fb81645562f7e495a49561065912788af2d1f14f/Formula/vcodex.rb#L5) passing of "magic words" in formulae, if that works here. Otherwise, not aware of one sadly.

ilovezfs:
@DomT4 In this case it looks like the magic is to make it
`:pserver:anonymous:@curlftpfs.cvs.sourceforge.net:/cvsroot/curlftpfs", :using => :cvs`
instead of
`:pserver:anonymous@curlftpfs.cvs.sourceforge.net:/cvsroot/curlftpfs", :using => :cvs`, which it turns out is how all of our other cvs URLs already are.

DomT4:
Took me a minute to actually spot the difference there! Feel free to 🚢 when you're happy here.

PR #213, opened 2016-06-07
i3lock 2.8 (new formula)

afh:
i3lock is a simple screen locker like slock. After starting it, you will
see a white screen (you can configure the color/an image).
You can return to your screen by entering your password.

### DIFF
--- a/i3lock.rb
+++ b/i3lock.rb
@@ -1,0 +1,33 @@
+class I3lock < Formula
+  desc "Simple screen locker"
+  homepage "http://i3wm.org/i3lock"
+  url "http://i3wm.org/i3lock/i3lock-2.8.tar.bz2"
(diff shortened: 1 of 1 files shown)


apjanke:
@BrewTestBot test this please

apjanke:
@BrewTestBot test this please

MikeMcQuaid:
Passing on this but we may consider it in homebrew/core (as this tap is being moved back there).

afh:
@MikeMcQuaid Ok, thanks for the update. It has become of little importance to me, since I don't care much about homebrew anymore, since it has moved into a direction that I disagree with.

Wishing you much success with the project and thanks for all the effort you and other folks have put into it! 👋  🙂

PR #239, opened 2016-08-12
adding freerdp stable-1.1 as devel

untoldone:


untoldone:
fixing to better match best practices from https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md

PR #243, opened 2016-09-01
yarp: Removed yarp formula

francesco-romano:
Yarp formula is now provided by the following tap:
[robotology/cask](https://github.com/robotology/homebrew-cask)


apjanke:
Merged!

Issue #248, opened 2016-09-21
openmotif formula


oliviersoares:
I'm getting this error when installing openmotif:

```
==> ./autogen.sh --prefix=/usr/local/Cellar/openmotif/2.3.4 --disable-dependency-tracking --disable-silent-rules --disable-jpeg --disable-png
==> make install
Error: No such file or directory - (/usr/local/Cellar/openmotif/2.3.4/share/man/man3/CORE.3, /usr/local/Cellar/openmotif/2.3.4/share/man/man3/openmotif-CORE.3)
```

Seems like the file:
https://github.com/Homebrew/homebrew-x11/blob/master/openmotif.rb#L62
is missing (/usr/local/Cellar/openmotif dir is missing).

I'm on macOS 10.12. Was able to install on 10.11.
Thanks.

apjanke:
This may have been fixed by #250. Could you `brew update` and try installing again?

oliviersoares:
@apjanke yes it seems fixed now thanks.

oliv@oliv:~$ ll /usr/local/Cellar/openmotif/2.3.6/lib/libXm*
-r--r--r--  1 oliv  staff   2.2M Oct  5 07:47 /usr/local/Cellar/openmotif/2.3.6/lib/libXm.4.dylib
-r--r--r--  1 oliv  staff   3.3M Jun 10 19:06 /usr/local/Cellar/openmotif/2.3.6/lib/libXm.a
lrwxr-xr-x  1 oliv  staff    13B Jun 10 19:06 /usr/local/Cellar/openmotif/2.3.6/lib/libXm.dylib@ -> libXm.4.dylib

However, when i try to install nedit (which is the one I was installing in the first place, openmotif coming as a dependency), it seems to not find the library when trying to compile it.

Here's the full log:

==> Installing nedit from homebrew/x11
==> Downloading https://downloads.sourceforge.net/project/nedit/nedit-source/nedit-5.6a-src.tar.gz
Already downloaded: /Users/oliv/Library/Caches/Homebrew/nedit-5.6a.tar.gz
==> Patching
patching file util/motif.c
==> make macosx MOTIFLINK='-lXm'
Last 15 lines from /Users/oliv/Library/Logs/Homebrew/nedit/01.make:
OpenMotif 2.3
OpenMotif is available from:
	http://www.opengroup.org/openmotif/
	http://www.ist.co.uk/DOWNLOADS/motif_download.html

[...]

PR #250, opened 2016-09-30
openmotif: Fix case of man page

jasonrm:
Failed when on a case sensitive filesystem.

Likely the cause of #248


apjanke:
Looks good. Merged. Thank you for your contribution to Homebrew!

PR #252, opened 2016-10-03
openmotif 2.3.6

apjanke:


apjanke:
It looks like this broke the builds for `nedit` and possibly `grace`. See #248 and https://github.com/Homebrew/homebrew-x11/issues/255#issuecomment-253087894.

Reverting in PR https://github.com/Homebrew/homebrew-x11/pull/260.

PR #256, opened 2016-10-07
nedit: allow openmotif 2.3.6

### LINKED ISSUE #248: openmotif formula

apjanke:
Upgrading to openmotif 2.3.6 broke the `nedit` build, because nedit checks for specific versions of motif that it is known to work with. Same thing happened with openmotif 2.3.4.

We need to either roll back to openmotif 2.3.4, or adjust our patch to nedit to allow 2.3.6. I'm submitting this PR for the latter, since that's how we handled it last time around.

Fixes #248.


apjanke:
Haven't heard back from nedit upstream on this yet. But it looks like OpenMotif 2.3.6 also broke the `grace` build. Putting in a PR to roll back to OpenMotif 2.3.4 (https://github.com/Homebrew/homebrew-x11/pull/260) which will get `nedit` building again, too, so merging this is not urgent.

apjanke:
@BrewTestBot test this please

apjanke:
@BrewTestBot test this please

apjanke:
Haven't heard from upstream yet; I think we should go ahead and do this (since I think we're sticking with OpenMotif 2.3.6), and deal with any issues if/when they arise.

apjanke:
Merged.

PR #260, opened 2016-10-12
openmotif 2.3.4 (revert 2.3.6)

### LINKED ISSUE #248: openmotif formula

apjanke:
This reverts commit 1b95bbf4c896e4a39f746df9b8f1e25653a64bcd.

Looks like openmotif 2.3.6 breaks the build for nedit and [grace](https://github.com/Homebrew/homebrew-x11/issues/255#issuecomment-253087894). There's a PR #256 open to make nedit build with openmotif 2.3.6 but not clear if that's supported by upstream, and getting grace to build with it needs some investigation. Let's roll back to 2.3.4 until we know how to get all its deps to build cleanly with 2.3.6 (as opposed to just getting the bottles or previously-built installations to run against it).

Fixes #248.

### DIFF
--- a/openmotif.rb
+++ b/openmotif.rb
@@ -1,13 +1,14 @@
 class Openmotif < Formula
   desc "LGPL release of the Motif toolkit"
   homepage "https://motif.ics.com/motif"
-  url "https://downloads.sourceforge.net/project/motif/Motif%202.3.6%20Source%20Code/motif-2.3.6.tar.gz"
-  sha256 "fa810e6bedeca0f5a2eb8216f42129bcf6bd23919068d433e386b7bfc05d58cf"
+  url "https://downloads.sourceforge.net/project/motif/Motif%202.3.4%20Source%20Code/motif-2.3.4-src.tgz"
+  sha256 "637efa09608e0b8f93465dbeb7c92e58ebb14c4bc1b488040eb79a65af3efbe0"
 
   bottle do
-    sha256 "9c90a53591bfe54c2c336c08795aeeb6182b804cbb5db293e8a6268c1d321535" => :sierra
-    sha256 "a08b6eb674784b04b034658bf078b31cb2a2145256ccf9508c4fcf1e1bfd5124" => :el_capitan
-    sha256 "b0158bbe9af692f60dda0c764d1c32da0c57640fc77be2fe37f4657c0ff2030d" => :yosemite
+    revision 2
+    sha256 "a7306b58b4f448614f89ede7bbf21e6d9f8618268ff99d039ecfc0473b29e3f3" => :el_capitan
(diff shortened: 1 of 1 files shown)


ilovezfs:
It doesn't look like this actually fixes nedit. Downgrading to Xquartz 2.7.8 and rebuilding openmotif 2.3.6 does fix it though.

apjanke:
I think the Xquartz 2.7.8 might be a separate issue, at least for `nedit`. `nedit` is checking for specific versions of openmotif in its build process, and refuses to build against other versions. Downgrading to Xquartz 2.7.8 fixes #255 (due to the two-level namespace linkage issue). But this fix rollback is for an independent issue #248 with the `nedit` build.

Note that this PR is specifically for a build time issue, not a run time issue. To get a fully-working `nedit` that can both build and run, I think we need to both do this PR (or patch the `nedit` build script) (to fix the build failure) *and* do the downgrade to Xquartz 2.7.8 like you mention (to fix the linkage issue).

I don't remember why this broke the build for `grace`; I may have had it confused with #255.

ilovezfs:
Why not just hack the nedit Makefile to allow it? It seems to build and run fine at that point.

apjanke:
> Why not just hack the nedit Makefile to allow it? It seems to build and run fine at that point.

Just because upstream is conservative about which OpenMotifs they support and have said they've had trouble with several releases in the past, and I'm not able to do thorough testing on it.

There's actually a PR open to do that Makefile hack: #256. Haven't heard back from upstream about it yet.

The [grace build issue](https://github.com/Homebrew/homebrew-x11/issues/255#issuecomment-253087894) looks like it's a similar strict version-checking failure; it could be hacked similarly.

apjanke:
@ilovezfs If you want to get the ball rolling on hacking the nedit Makefile, do you mind reviewing and merging #256? Then I'll make an upstream bug report for the grace build failure once we're done with the `qt5` openssl stuff and I have some available time.

apjanke:
Closing this out since it looks like we have a way forward with OpenMotif 2.3.6, with #265.

PR #261, opened 2016-10-12
Add fix for 'dyld: lazy symbol binding failed' to freeglut

Git-Jiro:
I just copied over and tested this fix here: https://github.com/Homebrew/homebrew-core/commit/964f9dc802a10475eba3021aa2a1dd329419f648


ilovezfs:
This is set in superenv so should not be necessary unless your brew is outdated.

PR #265, opened 2016-10-25
openmotif: fix VendorShell reference for XQuartz 2.7.9 and 2.7.10

### LINKED ISSUE #255: openmotif (grace, ddd, nedit, xpdf) broken with XQuartz 2.7.9, 2.7.10
There's a linkage issue with openmotif and the post-2.7.8 versions of XQuartz. XQuartz changed the linkage options for `libXt` to use a two-level namespace, which openmotif is currently incompatible with. It requires a flat namespace dylib. The openmotif build will seem to succeed, but when you run applications that use openmotif, they will crash on startup with errors like the following:

```
Warning: Widget must be a VendorShell. Warning: Fatal Error: _XmGetDefaultDisplay cannot be used prior to VendorS.Initialize, returns NULL

Oops! Got SIGSYS

Please use "Help/Comments" to report the bug. Abort trap: 6
```

Caused https://github.com/Homebrew/brew/issues/1189.

It looks like the bottles for at least some OS X versions are broken because they were built against newer XQuartz distributions.

Affected programs:
- `homebrew/science/mbsystem`
- `homebrew/science/xmgredit`
- `homebrew/x11/ddd`
- `homebrew/x11/geomview`
- `homebrew/x11/grace`
- `homebrew/x11/nedit`
- `homebrew/x11/xastir`
- `homebrew/x11/xpdf`
## Workaround

As a workaround, to get motif-based programs running again:
- Downgrade XQuartz to 2.7.8
- `brew update && brew upgrade`
- Re-install `openmotif` and the affected programs from source
## Analysis

There are workarounds available with XQuartz 2.7.9 and 2.7.10. In 2.7.9, `libXt.6.dylib` is linked with a flat namespace, while `libXt.7.dylib` has the new two-level namespace. In 2.7.10, they changed their mind, and now both `libXt.6.dylib` and `libXt.7.dylib` are linked with two-level namespaces, but flat namespace versions of them are provided in a separate directory.

From the [2.7.10_rc4 release notes](https://www.xquartz.org/releases/XQuartz-2.7.10_rc4.html):

> - libXt
>   - libXt.6.dylib is now a two-level-namespace dylib
>   - libXt.7.dylib is now a stub dylib that re-exports libXt.6.dylib (for binary compatibility with anything built against XQuartz 2.7.9)
>   - A flat_namespace version of libXt is available in /opt/X11/lib/flat_namespace to help ease the transition (#96292)
>     - Set DYLD_LIBRARY_PATH=/opt/X11/lib/flat_namespace when executing older non-compliant software (eg: Motif-based applications)
>     - Motif users are encouraged to file bugs against Motif to encourage them to fix that library.

So the build would need to be done differently depending on which version of XQuartz is installed. To get the build to work, we need to get it linking against the flat-namespace version of `libXt`.

This means that we can't provide bottles for `openmotif` unless we require a specific version of XQuartz. (At least for the bottles, and we don't have a mechanism for tying bottles to specific versions, unless we just do a procedural check in the `pour_bottle?` test.) And that version of XQuartz would need to be matched to the one on the bottle-building test-bot servers.
## What to do

The bottles are probably currently broken, so we should probably disable the bottles for `openmotif` (and possibly the programs using it?), add a note to its caveats that it requires XQuartz 2.7.8 or earlier, and add a max-version test to the `:x11` requirement and enable it for `openmotif`.

We could probably fix the bottles by downgrading XQuartz on the test-bot servers to 2.7.8 and rebuilding the bottles.

Then figure out how to get it linked against the flat-namespace dylibs from XQuartz 2.7.9 and 2.7.10. At that point, maybe re-upgrade XQuartz on the test-bot servers and make bottle pouring conditional on that version. It would be simpler to just disable the bottle for `openmotif`, though. It's not a long build, so it wouldn't hurt users much to have them build it from source.

Make sure there's an upstream bug on OpenMotif about this, and encourage them to fix it.

bhelf:
The XQuartz 2.7.9 and 2.7.10 releases changed the libXt dylib to be built without -flat_namespace.  This breaks Xt-based widget libraries that overload VendorShell by using ELF-style or static-linking-style tricks that expect to be able to replace a symbol just by using it in a "higher-level" library.  

This change is based on the change that the XQuartz maintainer did to libXaw, which does the same trick that Motif does.

It lets nedit and grace run again with XQuartz 2.7.9 and 2.7.10.

This fixes #255

### DIFF
--- a/openmotif.rb
+++ b/openmotif.rb
@@ -121,3 +121,44 @@ index e3f56bd..d056e03 100644
  #include <Xm/Xm.h>
  #include "wsm.h"
  #include "wsmDebug.h"
+
+
+diff --git a/lib/Xm/VendorS.c b/lib/Xm/VendorS.c
+index b5b5a24..cbd0967 100644
+--- a/lib/Xm/VendorS.c
++++ b/lib/Xm/VendorS.c
+@@ -292,6 +292,25 @@ static unsigned short destroy_list_cnt ;
+ static Display * _XmDisplayHandle = NULL ;
+ static XtErrorMsgHandler previousWarningHandler = NULL;
+ 
++
++
++#if defined(__APPLE__)
++/* Hack necessary to handle Apple two-level namespaces */
++extern WidgetClass vendorShellWidgetClass; /* from Xt/Vendor.c */
++extern VendorShellClassRec _XmVendorShellClassRec;
++#define vendorShellClassRec _XmVendorShellClassRec
++
++__attribute__((constructor))
++static void __VendorShellHack(void)
++{
++    vendorShellWidgetClass = (WidgetClass)(&_XmVendorShellClassRec);
++    transientShellWidgetClass->core_class.superclass = vendorShellWidgetClass;
++    topLevelShellWidgetClass->core_class.superclass = vendorShellWidgetClass;
++}
++#endif
++
++
++
+ /***************************************************************************
+  *
+  * Vendor shell class record
+@@ -506,6 +525,8 @@ VendorShellClassRec vendorShellClassRec = {
+     }                            
+ };	   
+ 
++#undef vendorShellClassRec
++
+ externaldef(vendorshellwidgetclass) WidgetClass 
+   vendorShellWidgetClass = (WidgetClass) (&vendorShellClassRec);
+ 


henrybaxter:
Thanks @bhelf

apjanke:
Nice find.

Would you mind submitting this upstream to OpenMotif itself? We don't need to wait for their approval to merge it, but we'd like to at least have a link to a report. Their bug tracker at http://bugs.motifzone.net/enter_bug.cgi is down, but you could still send a report or make a forum post through http://motif.ics.com/.

bhelf:
So, I tried to register for the ICS Motif forums, but their registration link appears to be broken.  And as you pointed out, their bug tracker is dead also.  The sourceforge page doesn't seem to have anything useful either, unfortunately.

I'm not entirely sure how to proceed -- I guess sourceforge has github-style fork/pull-request functionality now, so I guess I can go make an account and see if that works.

bhelf:
Also, I looked at the failing checks for this, and it looks like it's automake getting unhappy?  I'm not sure what's going on with that; the change doesn't touch anything involving the build system, so I'm not sure what would cause this.  Is there something that needs to be tweaked in the formula file to fix that?

apjanke:
> So, I tried to register for the ICS Motif forums, but their registration link appears to be broken...

I think we need to email them through their "Contact Us" page and establish a one-on-one communication channel. Besides this bug report, it'd be nice to get some clarity on their overall approach to the two-level libs, and release cycle. I'll do this, unless you're itching to volunteer.

apjanke:
> Also, I looked at the failing checks for this, and it looks like it's automake getting unhappy? ...

This looks more like a transient problem with the test-bot. It's not even properly picking up test output. Here's an excerpt from the console log (which in normal cases you do not need to dig down in to).

```
==> brew fetch --retry homebrew/x11/openmotif --build-bottle --force
==> brew install --only-dependencies --build-bottle --verbose homebrew/x11/openmotif
Agent went offline during the build
ERROR: Connection was broken: java.io.IOException: Unexpected termination of the channel
	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:73)
Caused by: java.io.EOFException
	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2353)
	at java.io.ObjectInputStream$BlockDataInputStream.readShort(ObjectInputStream.java:2822)
	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:804)
	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:301)
	at hudson.remoting.ObjectInputStreamEx.<init>(ObjectInputStreamEx.java:48)
	at hudson.remoting.AbstractSynchronousByteArrayCommandTransport.read(AbstractSynchronousByteArrayCommandTransport.java:34)
	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:59)

Build step 'Execute shell' marked build as failure
ERROR: Step ‘Publish JUnit test result report’ failed: no workspace for Homebrew X11 Pull Requests/version=el_capitan #500
ERROR: Step ‘Archive the artifacts’ failed: no workspace for Homebrew X11 Pull Requests/version=el_capitan #500
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
ERROR: Step ‘Delete workspace when build is done’ failed: no workspace for Homebrew X11 Pull Requests/version=el_capitan #500
Finished: FAILURE
```

Let's see if this is truly transient and has resolved itself on its own.

bhelf:
Oh, hey, looks like ICS has the Motif bugzilla back up and running again.  Woo!
I submitted this there: http://bugs.motifzone.net/show_bug.cgi?id=1655

bhelf:
I did also test this patch with XQuartz 2.7.11 and it continues to fix the problem.

apjanke:
@BrewTestBot test this please

apjanke:
I've tested it with XQuartz 2.7.8 (by building and running `nedit` and `grace` against it) for back-compatibility, and it looks good there too. I think it's time to proceed with this.

The formula should have a revision bump since it's changing the installed binaries. Could you add a `revision 1` to the formula after the main `sha256` line?

apjanke:
I merged this in, along with a revision bump, in #270. Should be fixed now.

PR #266, opened 2016-10-28
xcb-util-xrm 1.0 (new formula)

afh:
XCB utility functions for the X resource manager

Upcoming version of [i3](https://github.com/Homebrew/homebrew-x11/blob/master/i3.rb) (i.e. the current `next` branch) depend on this.

### DIFF
--- a/xcb-util-xrm.rb
+++ b/xcb-util-xrm.rb
@@ -1,0 +1,31 @@
+class XcbUtilXrm < Formula
+  desc "XCB utility functions for the X resource manager"
+  homepage "https://github.com/Airblader/xcb-util-xrm"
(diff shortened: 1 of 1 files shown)


ilovezfs:
I vendored this into the i3 PR so it shouldn't be needed: https://github.com/Homebrew/homebrew-x11/pull/268

@afh do you mind testing if #268 works for you?

afh:
@ilovezfs thank you for the work in #268 I just verified that it works very well for me! 👍

afh:
Closing in favour of #268.

ilovezfs:
@afh Thank you for testing that! I will ship it now. (If another formula ends up needing xcb-util-xrm at some point, we can make this a separate formula.)

PR #272, opened 2016-11-24
freeglut fix

nidegen:
Fix as hinted by @ilovezfs


apjanke:
Audit is failing.

```
        Error: 1 problem in 1 formula
homebrew/x11/freeglut:
  * cmake dependency should be
      depends_on "cmake" => :build
    Or if it is indeed a runtime dependency
      depends_on "cmake" => :run
```

Looks unrelated, but would you mind fixing it while you're in here?

dunn:
@BrewTestBot test this please

dunn:
Thanks for your patience!

PR #275, opened 2016-12-05
jskeus 1.0.13: fix sha256 for force pushed tag

wkentaro:
Close https://github.com/euslisp/jskeus/issues/403


dunn:
From that ticket:

> Maybe the tag is force pushed?

wkentaro:
> Did that turn out to be true?

Yes.

dunn:
Cool, thanks.

wkentaro:
Thanks.

PR #269, opened 2016-11-12
libxkbcommon 0.7.1

ilovezfs:
Created with `brew bump-formula-pr`.

PR #279, opened 2016-12-20
libxkbcommon: use secure urls

vszakats:


vszakats:
Anyone to merge this? The patch seems valid and correct.

vszakats:
There is a collision with open PR: https://github.com/Homebrew/homebrew-x11/pull/269
I've left comments there to sync it with this patch. Anyhow till that one gets sorted out, it'd be nice to merge this one.

PR #281, opened 2016-12-29
Delete unused formulae

MikeMcQuaid:
Delete all homebrew/x11 formulae that show no installations since analytics records began.


dancrossnyc:
Aw shucks; just when I wanted to install xastir. :-)

Issue #283, opened 2017-01-01
openmotif install fails with failed patch


smalyshev:
Get this when trying to install openmotif

```
==> Installing openmotif from homebrew/x11
==> Downloading https://downloads.sourceforge.net/project/motif/Motif%202.3.6%20Source%20Code/motif-2.3.6.tar.gz
Already downloaded: /Users/smalyshev/Library/Caches/Homebrew/openmotif-2.3.6.tar.gz
==> Patching
patching file configure.ac
patching file lib/Xm/XmP.h
patching file demos/programs/workspace/xrmLib.c
patching file lib/Xm/VendorS.c
Hunk #2 FAILED at 525.
1 out of 2 hunks FAILED -- saving rejects to file lib/Xm/VendorS.c.rej
Error: Failure while executing: /usr/bin/patch -g 0 -f -p1
```

dunn:
Do you know if that patch has been applied upstream, or if there's a refreshed patch we need to include?

smalyshev:
@dunn I don't know, the patch just fails, I don't know Motif build enough to see what's missing.

dunn:
Removing the patch causes the build to fail with the error reported here: http://bugs.motifzone.net/long_list.cgi?buglist=1625

Given that it's been left open for three and a half years without comment, I think the conclusion is that OpenMotif is not supported on macOS.

dunn:
@smalyshev you should be able to install it now!

smalyshev:
yes, no more patch failures, thank you!

PR #289, opened 2017-01-19
openmotif: fix patch

alyssais:
Some whitespace was missing.

Fixes homebrew/homebrew-core#9071.


ilovezfs:
Should probably use a patch in https://github.com/Homebrew/formula-patches.

dunn:
This PR should probably Close https://github.com/Homebrew/homebrew-x11/issues/283 as well.

It's worth thinking about how long we want to carry these patches without upstream support, also.

alyssais:
Oh, yeah, thanks @dunn. I didn't notice that.

PR #293, opened 2017-01-27
Remove final formulae.

MikeMcQuaid:
These weren't able to be migrated in the first batch because `curlftpfs` needs OSXFuse and homebrew/core CI can't handle that yet and `mesalib-glw` had to have options tweaked.


MikeMcQuaid:
Closing in favour of separate PRs in https://github.com/Homebrew/homebrew-x11/pull/294 and https://github.com/Homebrew/homebrew-x11/pull/295

